PR #50, opened 2017-08-07
Removed link to the Medium blog

pawelmadeja:
I think that we no longer need it since when going to the URL, there appears message: "The user deleted their Medium account" ;)


Atinux:
Thank you @pawelmadeja

PR #126, opened 2018-03-25
enhancement: re-use code for the menu dropdown. Remove flags from translation dropdown and use iso codes instead (closes #62)

### LINKED ISSUE #62: Languages should never be represented with a national flag
Languages should never be represented with a national flag.

1. A country with a history of British colonization may still use English as their official language, and using the Union Jack to denote their official language could be offensive.
2. A country may have more than one language in common use, and using the national flag to refer to only one language would mean ignoring the other communities.
3. Using the German national flag to refer to the German language would imply that the Federal Republic of Germany has some kind of language authority, while other German-speaking countries (Austria, Switzterland, etc.) don't.
4. Using national flags is not a scalable approach; Some languages don't have a corresponding flag at all.

Most languages are not one-to-one correspondence with a state. If you need a good way to identify languages, please use the name of them just like [vuejs.org](https://vuejs.org) does so.

Special thanks to @xtendo-org, for bringing up this problem.

<div align="right"><sub><em>This feature request is available on <a href="https://nuxtjs.cmty.io">Nuxt.js</a> community (<a href="https://nuxtjs.cmty.io/nuxt/nuxtjs.org/issues/c59">#c59</a>)</em></sub></div>

scramantha:


Atinux:
Thank you @scramantha 

Because of conflicts, I made the changes myself here: https://github.com/nuxt/nuxtjs.org/commit/8098452dc1d3a34679edff37c68741cbb12f9a94

PR #150, opened 2018-08-16
replace JWT sample code link.

konojunya:
reference: https://github.com/nuxt/docs/issues/710


konojunya:
Since download link of https://nuxtjs.org/examples/auth-external-jwt is wrong, I tried to fix it, but it seems that it was not here to send PR.
The community of nuxt was a bit complicated and I got lost. I'm sorry.

konojunya:
move to https://github.com/nuxt/nuxt.js/pull/3746

Issue #104, opened 2018-01-17
Convert images to SVG


patrickdaze:
Going forwards I was thinking of converting the PNG images to SVG for the following reasons:
* it can be used as both a source and a display file
* resizable/vector/smaller filesize
* it could be inserted inline for translations

**Upgrade path:**

1. Convert [nuxt-schema.png](https://github.com/nuxt/nuxtjs.org/blob/master/static/nuxt-schema.png) to SVG (keep PNG for now)
2. Convert [nuxt-views-schema.png](https://github.com/nuxt/nuxtjs.org/blob/master/static/nuxt-views-schema.png) to SVG (keep PNG for now)
3. Update [all translations](https://github.com/nuxt/docs) to point to SVG instead of PNG
4. Delete [nuxt-schema.png](https://github.com/nuxt/nuxtjs.org/blob/master/static/nuxt-schema.png) and [nuxt-views-schema.png](https://github.com/nuxt/nuxtjs.org/blob/master/static/nuxt-views-schema.png) once all translations points to the new SVG
5. (future/optional) Add SVG inline in translations (for multilingual support)

An example of the SVG file that could be used: [nuxt-schema-svg-example.zip](https://github.com/nuxt/nuxtjs.org/files/1640813/nuxt-schema-svg-example.zip)

Anyone think this would be beneficial?

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxtjs.org/issues/c99">#c99</a>)</em></sub></div>

Atinux:
I do!

PR #158, opened 2018-10-11
Convert schemas to SVGs

patrickdaze:
This PR includes two changes:

1. Converted [nuxt-schema.svg](https://github.com/patrickdaze/nuxtjs.org/blob/convert-to-svg/static/nuxt-schema.svg). It can be found on this page: https://nuxtjs.org/guide
2. Converted [nuxt-views-schema.svg](https://github.com/patrickdaze/nuxtjs.org/blob/convert-to-svg/static/nuxt-views-schema.svg). It can be found on this page: https://nuxtjs.org/guide/views

This PR does not update the text/content of the schemas. It only tries to replicate the same style and text as the original PNGs. The fonts used are the same as in [_document.scss](https://github.com/nuxt/nuxtjs.org/blob/master/assets/scss/base/_document.scss#L15).

Next Step: Once the PR is pulled-in and deployed, I'll create another PR on the nuxt-docs repo to update image links: [I started that here](https://github.com/nuxt/docs/compare/master...patrickdaze:convert-to-svg). I might also have to do a PR to individual translation repos (will investigate).

---

See Issue #104 or https://cmty.app/nuxt/nuxtjs.org/issues/c99


Atinux:
Sorry for the delay, thank you so much.

I will update the docs :)

PR #162, opened 2018-11-23
Fixed displaying translations of each country on dev mode.

tell-k:
Hi nuxtjs.org authors.

First, thank you for awesome documents.

When I ran this codes as dev mode(`npm run dev`), I coundn't watch the japanese transations. So I fixed the codes.

Since `dev mode` is `SSR`,  `setLocale` is never called.

Please feel free to merge it.
Thx :)


tell-k:
@alexchopin @Atinux 

Is there any problem with this PR?
If there are problems, I will fix it.
thx

Atinux:
Sorry for the delay @tell-k 

Thanks for this PR <3

PR #184, opened 2019-04-23
feat(static): Use full static mode and crawler for generation

Atinux:
Static version for production :)


modules/docs/plugin.js
```diff
@@ -0,0 +1,131 @@
...
+          this.returns.delete(obj.challenge)
+        }
+      break;
+      default:
+    }
+  }
+  async callMethod(name, ...args) {
+    if (!this.connected) {
+      // console.log('WS not connected, retrying in a sec...')
+      await new Promise((resolve) => setTimeout(resolve, 1000))
```
pi0: Is this needed?

PR #191, opened 2019-05-04
refactor: improve module

pi0:
- Server is rewritten into a reusable class: [server.js](https://github.com/nuxt/nuxtjs.org/blob/refactor/improve-module/modules/docs/server.js)
- Module simplified: [index.js](https://github.com/nuxt/nuxtjs.org/blob/refactor/improve-module/modules/docs/index.js)
- General refactor and improvements

TODO: 
- Speed-up generate
- Try regex approach

### DIFF
--- a/modules/docs/index.js
+++ b/modules/docs/index.js
@@ -67,10 +67,15 @@ module.exports = async function (moduleOptions) {
     })
 
     // Profile generate
-    const startTime = new Date()
-    let count = 0
+    let startTime
+    let count
     this.nuxt.hook('generate:routeCreated', () => {
-      count++
+      if (!startTime) {
+        startTime = new Date()
+        count = 0
+      } else {
+        count++
+      }
     })
     this.nuxt.hook('generate:done', () => {
       const time = (new Date() - startTime) / 1000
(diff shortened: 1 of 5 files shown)


modules/docs/index.js
```diff
@@ -1,52 +1,46 @@
...
-  })
 
-  this.nuxt.hook('listen', () => startServer(join(this.options.srcDir, 'docs'), port))
-  // We need to run the server in order to serve the docs when running `nuxt generate`
-  this.nuxt.hook('generate:before', async (generator) => {
-    let server
+  // Start docs server
+  if (runServer) {
+    const docsServer = new DocsServer(options)
+    await docsServer.cloneRepo()
```
Atinux: We only want to clone the repo after the production build and before generating.

In development, I prefer them to clone `nuxt/docs` directly so they can directly push when working on the docs :)

See https://github.com/nuxt/nuxtjs.org#setup

pi0: Yes. This is handled inside `cloneRepo` method. When target dir exists it will ignore cloning.

PR #200, opened 2019-06-05
Make anchor icon visible on focus

kmelve:
Just a small accessibility fix to add visibility to the anchor icon on `:focus`. 

Before:
image

After:
image


Atinux:
Thanks @kmelve :)

PR #240, opened 2019-09-30
Fix: heading underlines

JasonTame:
### Context:
The horizontal lines below headings in the docs are floating in the middle of the text if the headings span multiple lines:

nuxtjs_headings_issue
 

### Change(s):

I've changed the `::before` pseudo-element added to each heading level to be `::after` and made it a block element with a border colour matching the background colour it had before. It will now always be below the text

nuxtjs_headings_fix


Atinux:
Awesome thank you @JasonTame

PR #244, opened 2019-10-02
Make text more readable on mobile + shape-outside for round SVG

mornir:
On mobile (tablet and desktop are fine), when the text is justified, it sometimes leaves big gaps:
image

We could also center align it, but I feel that a left aligned text is more readable.

As for the text about sponsoring, I suggest having it always left aligned.

PS: I love the redesign with Tailwind 😍


mornir:
I also added the `shape-outside` property to the rounded SVG. I think the effect looks nice 😎
I added some margin below the h1 "Sponsors" on mobile, because the SVG was too close to the title.

PR #303, opened 2020-02-29
feat: dark mode

bdrtsky:
Dark/light theme switcher. Slightly different and simpler implementation than were discussed with @Atinux 

Also minimal CSS changes for demonstration purpose, but in general CSS should be changed a lot to support theming. But first design and design tokens should be discussed and approved.


bdrtsky:
Looks like https://github.com/nuxt-community/universal-storage-module overwrites local session data in production. In dev it doesn't rewrite. Not sure what is happening here.

bdrtsky:
Follow up - https://github.com/nuxt-community/universal-storage-module/issues/75

components/partials/Footer.vue
```diff
@@ -94,10 +95,12 @@ export default {
   computed: {
     currentTheme: {
       get () {
-        return this.themes.map(l => l.value).indexOf(this.$store.state.storage.theme)
+        const index = this.themes.map(l => l.value).indexOf(this.$theme.value)
```
pi0: - What about `const index = this.themes.findIndex(l => l.value === this.$theme.value) `?
- Statically generated version focuses on the default selected item until the client-side is being loaded. Not a problem for us because the button is in footer out of view and also no hydration because the markup is the same just attr change happening. Just caution this problem is still there for general usage:

image

modules/theme/plugin.js
```diff
@@ -0,0 +1,44 @@
+import Vue from 'vue'
+import { parse, serialize } from 'cookie'
+
+const key = '<%= options.cookie.key %>'
+const cookieOptions = <%= JSON.stringify(options.cookie.options, null, 2) %>
+
+export default function (ctx, inject) {
+  const theme = Vue.observable({
+    value: (process.client ? window.__nuxt_theme : '<%= options.value %>')
```
pi0: - Can we just make `value` observable? Assigning a function to a reactive object is little bit hacky and confusing when one tries to debug `$nuxt.$theme`
- The field name of `value` for us makes sense but probably little bit confusing for a general variable naming. What about `$theme.name` or `$theme.variant` (also same for nuxt config)

modules/theme/plugin.js
```diff
@@ -0,0 +1,44 @@
...
+  })
+  const setTheme = (value, persist = true) => {
```
pi0: I would suggest `{ persist }` so in future we can easily extend it :)

modules/theme/plugin.js
```diff
@@ -0,0 +1,44 @@
...
+    theme.value = value
+    if (process.client) {
```
pi0: Maybe a todo for future improvements. One may call `$theme` on the server-side which we will lose it in this case. We can use `res` to set cookie :)

modules/theme/plugin.js
```diff
@@ -0,0 +1,44 @@
...
+      // update cookie
+      if (persist) {
+        document.cookie = serialize(key, value, cookieOptions)
```
pi0: Is it really required to use an external package (800B - 16ms) to serialize a string field? We can use a string template and also consider router.base for theme scope.

modules/theme/plugin.js
```diff
@@ -0,0 +1,44 @@
...
+
+  // Reads from color-scheme
+  if (process.client && window.matchMedia) {
+    const darkQuery = window.matchMedia('(prefers-color-scheme: dark)')
+    darkQuery.addListener((e) => setTheme(e.matches ? 'dark' : 'light', false))
+  }
+
+  if (process.static && process.client) {
+    // Static generated, read from cookie + update body attrs manually
+    document.body.setAttribute('data-theme', theme.value)
```
pi0: Why not resue `setTheme` or create a shared fn for `body.setAttr`?

modules/theme/script.js
```diff
@@ -0,0 +1,23 @@
+// Add dark / light detection that runs before Nuxt.js load.
```
pi0: - As proposed in Notion, we can minify this injected script into a tiny IIFE
- `getCookie` can be reused by the plugin to save some bytes, mem and also more consistency (like `__nuxt_theme._getCookie`)
- Also other logic like query matching and setAttr is duplicated (not a big deal but still perf improvement area)

modules/theme/plugin.js
bdrtsky: Or `theme.mode`. I actually call this `color modes` in my apps, since `theme` for me is more like a template. But it's purely naming issue.

bdrtsky:
Didn't finish this blocks, since they require more designer input: Alert, Promote, Carbon, code blocks.
Also, some interactive elements, that have hover states, like Contributor Tag.

And issue with images, I am not sure how to proceed them. They are used all over the website as src in img tags, and they look bad in dark mode. 
Also, problem with hovered SVG's on pages like https://nuxtjs.org/resources
To fix this SVG should be cleaned up, and certain elements grouped, so I could change their fill.

components/partials/Footer.vue
Atinux: Well `findIndex` has bad IE support:
Screenshot 2020-03-09 at 11 29 39

bdrtsky:
Can someone take a look why after `generate` in dark mode icon is broken? It doesn't show right icon.

modules/theme/plugin.js
Atinux: Updated :)

modules/theme/plugin.js
Atinux: Future improvement yes 👍

Atinux:
Will merge now, when creating a community module for supporting dark mode on Nuxt websites, we will use @pi0 suggestions :)

components/partials/Footer.vue
pi0: It is already transpiled by babel :)

PR #323, opened 2020-03-31
Added Hive under "WHO'S USING NUXTJS"

therealwolf42:
Hive is a blockchain, where https://hive.io is the main website, which was coded via NuxtJS.

Repo is open-source: https://gitlab.syncad.com/hive/hive-io

This PR adds Hive under `WHO'S USING NUXTJS`. I've added a new row and set `justify-content: center` as it looked better than it being stuck on the left side.

image

I think the website (hive.io) is a good example on how to implement NuxtJS with Typescript & new Vue 3.0 Composition API.

PS: I love NuxtJS. Thanks for all the hard work you're all putting into this!


Atinux:
Thank you for the PR, actually we are in a process to update this to have a database where people can submit their website, we will keep you updated :)

PR #334, opened 2020-04-07
Align tags for small screens

viniciuskneves:
Currently this is what happens on small screens:
image

I've implemented the following solution which adds a wrapper div:
image

Another option (let me know what you think) is this one which is just CSS:
image


bdrtsky:
Screenshot 2020-04-07 at 21 03 17

Looks much better now, thanks!

Issue #349, opened 2020-04-22
Navbar does not respect safe-area-inset and overlap iPhone X Home Indicator 


mrleblanc101:
### What problem does this feature solve?

This would improve the mobile experience on iPhone X and newer as currently the bottom Navbar does not respect safe-area-inset for the Home Indicator

### What does the proposed changes look like?

Adding a padding-bottom: env(safe-area-inset-bottom) would prevent the Home Indicator from overlapping the Navbar.

![4B8FD155-028D-4C67-9676-BC7DC1A2978A.jpeg](https://cmty.s3.eu-west-1.amazonaws.com/73294699/1e314120-8461-11ea-9054-91b04ed0aea6.jpeg)

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxtjs.org/issues/c214">#c214</a>)</em></sub></div>

mrleblanc101:
I can work on this if someone gives me the go ahead. Shouldn't be too hard to fix.

Geminii:
Hi @mrleblanc101 
You can purpose a PR for this fix :) 
Just be careful about potential regression on other device.

mrleblanc101:
There you go #351

bdrtsky:
Where is exactly the issue on image? We already had this before, but it was removed since it broke padding on every other device beside iPhone https://github.com/nuxt/nuxtjs.org/pull/299

mrleblanc101:
@bdrtsky The issue it that the Navbar should not overlap the home indicator. That's why the same area inset API was created. 

![image](https://user-images.githubusercontent.com/17392251/80368791-49d02280-885b-11ea-884b-099400d833d3.png)

The old solution did not work as you added padding in a fixed height container, which reduced the height of the content. 
Ex: Container is 64px high. As you scroll, the safe-area-inset increase from 0 to  20px reducing the content height. An other solution than removing the fixed height would be to use `box-sizing: content-box;` instead of `border-box`

PR #351, opened 2020-04-23
fix: home indicator overlap on iphone x (#349)

mrleblanc101:


mrleblanc101:
Hi,

I used what was commented as a base.
Instead of putting the padding on the link themself, I added it to the Navbar as it was a cleaner way to do it. 

I had to remove the `h-16` class so it wouldn't interfere with the padding. It was not necessary as the nav will take the height of the biggest item which was also `4rem`.

This shouldn't cause any issue as the rule will be ignored on device that do not support `env()` and fallback to the current behavior. It will also be ignored on device which support `env()` but have no need for it because the variable value will be `0`.

mrleblanc101:
Simulator Screen Shot - iPhone 11 Pro - 2020-04-22 at 22 37 23

bdrtsky:
Why fixed height was removed?

mrleblanc101:
Because you can't have a fixed height if the padding value change as you scroll unless you use `box-sizing: content-box;` otherwise the content overflow the `content-box` or get squished

mrleblanc101:
The height is not necessary as this is a flex container. The container will take the height of the biggest element + it's padding.

bdrtsky:
Still, right now it's hieght is 60.81px, not 64px - https://deploy-preview-351--nuxt.netlify.app/

bdrtsky:
> The height is not necessary as this is a flex container. The container will take the height of the biggest element + it's padding.

It is neccesery, since it depends on footer padding value.

mrleblanc101:
It make no sense to use height in the first place. It should be min-height because it also break the layout of some text wrap on a second line on another language. Is there a min-height class version of the previous class ? Otherwise, I will add it in the CSS if those 3px are that important

mrleblanc101:
Here is what happen when you use fixed height:
Capture d’écran, le 2020-04-27 à 08 14 33

bdrtsky:
It shouldn't be min-height since 3 lines will be too much, and the better solution is to truncate lines...

Height makes perfect sense since like I told - this fixed nav sizing directly connected to footer. With your solution it's broken on every other device beside iPhone (you can see a line between top of fixed nav and bottom of the footer - it's body background appears since there's a gap appears because footer padding doesn't equal fixed nav height)

bdrtsky:
In general I agree, we need `env(safe-area-inset-bottom)` but without breaking padding/height sizing. They should be the same

bdrtsky:
What we could try - is to move `h-16 ` to nav items. Can you try it @mrleblanc101 on your iPhone?

mrleblanc101:
I'll look into the footer issue. It use `padding-bottom: 4em` which should be `4em + safe-area-inset-bottom`, otherwise the buttons are partially blocked

bdrtsky:
I think we can live your changes and move `h-16` to nav items, will it work for you?

mrleblanc101:
Yes, I'll update the PR tonight with both change (footer fix and moving the height to the element)

bdrtsky:
Also, what if use this on footer?

```
padding-left: env(safe-area-inset-left);
padding-right: env(safe-area-inset-left);
```

mrleblanc101:
I think you posted and deleted a comment about inset-left and inset-right, I could look into this for landscape I guess even through i doubt people will read documentation in landscape

bdrtsky:
@mrleblanc101 yes, i wasn't sure about it, because I can't properly test it right now. If you can test, then I believe you can add what you think is right, the main thing is preserve padding/height dimentions.

mrleblanc101:
Hi,

I've updated to PR to put the fixed height back and fix the partially blocked buttons in the footer (4em + safe-area-inset+bottom).

There was nothing to change in landscape for `safe-area-inset-left` and `safe-area-inset-right` as the meta `viewport-fit: cover` is not defined.

Landscape is not 100% perfect, but it wasn't before. I couldn't find a way to make it better because of some weird safari behavior with fixed element when the browser is in `reduced UI mode`, so I've added a media query to keep the current behavior.

Current:

If we were to apply the changes in landscape too:

bdrtsky:
Still missing fixed height on nav items, and now we have this on desktop

Screenshot 2020-04-29 at 11 10 25

mrleblanc101:
Sorry, I'm new to Tailwind, I've added a media query so the extra padding is only added under 768px. Look like `.h-full` was overwriting `.h-16` so I only kept one of the classes.
Do you have an idea why all the checks are failing? The log are confusing, talking about some yml file I did not touch.

lamalamaMark:
Hi guys, without knowing that this PR already exists I made another one tonight, see https://github.com/nuxt/nuxtjs.org/pull/354

I think the problem is solved in mine, but open for feedback. Adding the viewport-fit=cover attribute is necessary I believe.

mrleblanc101:
Viewport-fit is not necessary. In fact it break landscape if you add it without adding the necessary env() everywhere

bdrtsky:
closes #349

PR #354, opened 2020-05-01
Improve menu on iPhone X / 11

lamalamaMark:
By removing the fixed height of the nav element and adding padding-bottom via constant/env values we have correct functioning spacing on iPhone X / 11 after scroll. 
Adding the 'viewport-fit=cover' value to the viewport meta tag makes the padding work.


lamalamaMark:
🔥📱

bdrtsky:
fixed in #351

Issue #365, opened 2020-05-24
Route transition delay


noname202:
### Version

[v1.1.0](https://github.com/nuxtjs.org/releases/tag/v1.1.0)

### Reproduction link

[https://nuxtjs.org/](https://nuxtjs.org/)

### Steps to reproduce

Navigate the nuxtjs.org website.

### What is expected ?

Responsive user experience and almost no perceivable delays.

### What is actually happening?

When navigating the official NUXTJS website, page transitions feel kind of sluggish and there is a perceivable delay until the new page is displayed. This is especially noticable when compared to simmilar frameworks official pages such as Next.js, I had a look at the Source and it turns out that for some reason a 0.3 s CSS transition is being applied to the body element of the website. However, as far as I'm concerned, there is no visible transition animation.

```<body class="font-sans font-medium bg-light-surface dark:bg-dark-surface text-light-onSurfacePrimary dark:text-dark-onSurfacePrimary transition-colors duration-300 ease-linear ">```

```
.duration-300 {
    transition-duration: .3s;
}

.ease-linear {
    transition-timing-function: linear;
}
```

### Additional comments?

I'm not entirely sure if the delay is applied on purpose, but from my perspective it dramatically lowers the perceived performance.

Tested on Chrome Version 81 (Windows) and Safari (IOS 13.5)

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxtjs.org/issues/c217">#c217</a>)</em></sub></div>

bdrtsky:
What exactly does this mean "page transitions feel kind of sluggish"?

noname202:
> What exactly does this mean "page transitions feel kind of sluggish"?

Sorry for not being more precise.  Currently, the tailwind utility classes `transition-colors`, `ease-linear`, and most importantly `duration-300`  are used for the `body `element as well as each page's `div `container. While this has (at least from what I can judge) no visible effect on route / page transitions, the 0.3 s transition duration is still applied, effectively delaying route / page transitions by 0.3 s. This results in navigating the page feeling not as fast as it should be.  

[...]

PR #366, opened 2020-05-25
fix: route transition delay

noname202:
Fixes route transition delay described in issue #365 .


bdrtsky:
Can't reproduce this issue locally. Weird. Will look into it more.

PR #380, opened 2020-06-30
added 'Edit this page on Github' to blog pages.

hecktarzuli:
1) Added 'Edit this page on Github' on blog pages.
2) The AppContribute comp should probably be tweaked long-term as it contains things like page links that seem to make it less reusable.
3) The phrase 'want to contribute to the documentation' is a little weird for blog pages, but 'meh' for now.

### DIFF
--- a/components/common/AppContribute.vue
+++ b/components/common/AppContribute.vue
@@ -38,6 +38,7 @@ export default {
       return '/' + this.$route.params.section
     },
     list () {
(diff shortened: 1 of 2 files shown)


debs-obrien:
Thanks Josh. Actually we are launching the new docs next week and blog will then be using content module so for now we can leave this as is and will add to new one. But this is great feedback. Thanks

PR #395, opened 2020-07-22
fix(home): title is shown twice

manniL:
image

---

Alternatively (instead of deleting) we could fill that part with a paragraph again.

image


debs-obrien:
It should be the paragraph.  Good spot. It must have got lost when moving to i18n

manniL:
Updated it to use https://github.com/nuxt/nuxtjs.org/blob/dev/i18n/en-US.js#L73 😋

Issue #402, opened 2020-07-26
[dev] Nuxt.js vs NuxtJS vs Nuxt


manniL:
Hey all 👋 

at the moment, the terms **Nuxt.js**, **Nuxt** and **NuxtJS** are not used consistently, nor they are "explained" somewhere.

From my current understanding, these terms are used like that:

* The *framework* = **Nuxt.js**
* The *company* = **NuxtJS**
* "Colloqiually" referrencing to the Framework and ecosystem (should be avoided) = **Nuxt**

---

## Questions

1. Is this still correct?
2. Because many people (still) think that **NuxtJS** === **Nuxt.js**, could we somewhere explain the terms correctly?
3. Would you accept PRs to fix the terminology in case 1. is answered with "yes"?

<div align="right"><sub><em>This question is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxtjs.org/issues/c231">#c231</a>)</em></sub></div>

danielroe:
That's my read too. If `NuxtJS` is used for the company, I think it needs a little `?` icon or `<abbr>` popup that explains this.

debs-obrien:
Nuxt.js for framework and NuxtJS for company, correct. There were 2 instances in docs that had NuxtJS instead of Nuxt.js and that is now corrected. Nuxt can be used throughout documentation as it is explaining features, commands etc so using Nuxt is ok as we wouldn't say the Nuxt.js generate command as the Nuxt generate command sounds better here. Also Nuxt server etc. But referencing the framework for example with Nuxt.js you can put your page in a pages folder.... I like to think of it as as if I was writing Nuxt.js it would be the same as writing The Nuxt.js framework so to go with my example above it doesn't make sense to say The Nuxt.js framework generate command. So throughout the documentation that is what I have used so it should be consistent however if you find a place where it isn't then feel free to correct it. The new site has not been done and we have only worked on the guides section of the website and not any other so there will be inconsistencies with the old content which will be worked on when new site is released. But for sure a lot of the plugins/modules might need updating. 

Regarding question 2 I will talk with Alex about that and where it makes sense to do that.

debs-obrien:
We will have dedicated pages (discover section) in the new website to explain how the NuxtJS company was created with the success of the Nuxt.js framework.

manniL:
> Nuxt.js for framework and NuxtJS for company, correct.

Great 👍🏻

[...]

PR #403, opened 2020-07-27
 chore: use correct Nuxt.js/NuxtJS terminology 

manniL:
Related: #402 

Please double-check if I unintentionally changed sth. the wrong way


debs-obrien:
I think I will let @alexchopin review this as not sure if the intention is to change the naming across the whole site at this point. It might be best when we launch the new design as not sure if it matters changing titles and meta etc Idea for now is just to release new docs with the same look and feel as old site until new site is launched. Anyway this is a CEO decision so will leave it in Alex's hands.

manniL:
> I think I will let @alexchopin review this as not sure if the intention is to change the naming across the whole site at this point. 

For sure 😋
My intention was mostly to keep everything consistent, especially now that NuxtJS / Nuxt.js is used more and more often. I didn't see any drawback on the changes, e.g. it shouldn't cause any issues with meta description or titles besides now containing the right info & possibly avoid confusion for the reader 😄

alexchopin:
Thanks for your contribution @manniL however changes are not applicable everywhere and I can not merge this PR.
As we know we are working on the new documentation and this terminology issue will be fixed as soon as the new website will be released.

PR #406, opened 2020-07-27
chore: use actual country name

pimlie:
Holland is not a country, it should be The Netherlands :netherlands: :orange_heart: 

This is borderline blasphemy! :upside_down_face:


debs-obrien:
sorry my mistake.

PR #431, opened 2020-07-28
Fix typo: create nuxt-app -> create-nuxt-app

timpulver:
The command was misspelled and caused an error.


debs-obrien:
this is now already fixed :)

PR #434, opened 2020-07-28
add: i18n to pages

GuiguiWeb:
Add the ability to translate the different pages on the same model as the homepage.


debs-obrien:
nice work. that saved me a few hours 😃

PR #437, opened 2020-07-29
fix link to cached-components in the aside nav

chypriote:
Hello, 
I realized the sidebar on the current doc shows an item "Using nginx as a proxy" that links to the "How to cache Vue components?" page. This PR corrects it.


debs-obrien:
nice catch. thanks

PR #448, opened 2020-07-31
Add missing 'render' subkey in CSP-docs.

maikelvl:
The `render` subkey is missing in the docs.

This is a follow-up from nuxt/docs#2095
/cc @debs-obrien


debs-obrien:
thanks @maikelvl

PR #464, opened 2020-08-04
🔥 first structure of turkish translations

eggsywashere:


nuxt.config.js
```diff
@@ -190,6 +190,13 @@ export default {
         file: 'id-ID.js',
         name: 'Indonesian',
         domain: 'https://id.nuxtjs.org'
+      },
+      {
+        code: 'tr',
```
debs-obrien, line 195: can we comment out this for now like the spanish version as adding this will add a link to the lang selector in footer and we don't have a repo set up for turkish yet therefore this link will break. we are working on improving this but for now we need to hide this until we have a solution. thanks

pages/support.vue
```diff
@@ -18,7 +20,7 @@
             </template>
             <template v-slot:team>
               <NuxtLink to="/team" class="text-nuxt-green underline">
-                NuxtJS team
+                NuxtJS ekibi
```
debs-obrien, line 23: changing this here will change this for all languages. If we need to translate the word team then we need to modify this by adding this phrase to the i18n file and then calling it here. obviously we then need to adjust all other languages. I can do this if you like and then you can just translate the word team when it is done. so for now you can remove all changes to this file.

pages/support.vue
```diff
@@ -142,9 +144,9 @@ export default {
     TechnicalSupportIllustration
   },
   head() {
-    const title = 'The NuxtJS Support'
+    const title = 'NuxtJS Desteği'
```
debs-obrien, line 147: same here title and meta will be translated for all pages if we add it here therefore this needs to be done in i18n and not here

i18n/tr-TR.js
```diff
@@ -0,0 +1,305 @@
+module.exports = {
```
debs-obrien, line 1: great job 😃

pages/support.vue
eggsywashere: Wow, alright. Since it's not a big deal I'll just keep it default until you add those strings to i18n 🥰🙌

PR #470, opened 2020-08-05
docs(eslint-loader): use cache

iliyaZelenko:
This will fix performance in very large projects. For example, in our project, the `dev` mode (`yarn dev`) was launched without this for **30 minutes**, and after using the cache, for **50 seconds**!!! 

New version of https://github.com/nuxt/docs/pull/1846

### DIFF
--- a/content/en/guide/development-tools.md
+++ b/content/en/guide/development-tools.md
@@ -201,7 +201,10 @@ It is also recommended to enable ESLint hot reloading mode via webpack. This way
           enforce: "pre",
           test: /\.(js|vue)$/,
           loader: "eslint-loader",
-          exclude: /(node_modules)/
+          exclude: /(node_modules)/,
+          options: {
+            cache: true
+          }
         })
       }
     }


debs-obrien:
@Atinux or @pi0  can you check this and should we add this to new docs too and explain it better with info box or something?

pi0:
I think we should recommand [eslint-module](https://github.com/nuxt-community/eslint-module#cache) as does the same. `cache` is disabled by default maybe there are some known issues (/cc @ricardogobbosouza) but otherwise we can enable by default for module :)

ricardogobbosouza:
The `eslint-module` needs to be updated. Today, `eslint-loader` is used, which will be depreciated in favor of `eslint-webpack-plugin`.
In `eslint-webpack-plugin` the **cache** is handled directly by `eslint`. In the next version we can enable it by default, this should greatly reduce the time.

ricardogobbosouza:
https://github.com/nuxt-community/eslint-module/issues/38

debs-obrien:
Another thing to note is that we do not have such a page in the new docs. As it is not exactly nuxt related but eslint therefore we have not included it in the new guides. but perhaps it will feature in the resources page or similar. 

Perhaps a tips page too 🤔

ricardogobbosouza:
I think we can do that
#1057

debs-obrien:
ok cool lets close this and work on migrating docs for the module and add a recommendation in docs to use module. Thanks @ricardogobbosouza

PR #489, opened 2020-08-07
fix: BlogpostPreviewItem navigation

### LINKED ISSUE #423: Incorrect article link behaviour on nuxtjs.org/blog
### Version

[v1.1.0](https://github.com/nuxtjs.org/releases/tag/v1.1.0)

### Reproduction link

[nuxtjs.org/blog](nuxtjs.org/blog)

### Steps to reproduce

1. Go to nuxtjs.org/blog
2. Open any blog post with cmd + mouse click (Mac) or ctrl + mouse click (Windows).
3. Link opens in a new tab but also navigates to the article. Now 2 same articles are open.

### What is expected ?

When using cmd / ctrl with mouse click a link should open ONLY in new tab without affecting current page.

### What is actually happening?

cmd / ctrl with mouse click opens blog article twice.

### Additional comments?

- It seems that this problem exists only on /blog page.
- Clicking an article link with mouse wheel (middle click) works fine.

pierresaid:
This concern the /blog/ page.

Changed the `BlogpostPreviewItem.vue` root tag from `article` to `a`.
Removed the navigation on click.

fix #423

### DIFF
--- a/components/partials/blog/BlogpostPreviewItem.vue
+++ b/components/partials/blog/BlogpostPreviewItem.vue
@@ -1,19 +1,14 @@
 <template>
   <NuxtLink
-    tag="article"
     :to="localePath({ name: 'blog-slug', params: { slug: post.slug } })"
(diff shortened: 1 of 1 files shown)


debs-obrien:
thanks 😊

PR #492, opened 2020-08-10
en: add extra options example for extractCSS

tim-yao:
When using `extractCSS`, we also can pass `extract-css-chunks-webpack-plugin` options. Which is supported in https://github.com/nuxt/nuxt.js/blob/15bc5b06e69e08a2dea325dac18cac0dde6d9ca5/packages/webpack/src/config/base.js#L385

This is helpful as a lot users like me got `Conflicting order` warnings from `extract-css-chunks-webpack-plugin`. See issue https://github.com/nuxt/nuxt.js/issues/4885#issuecomment-533015230


debs-obrien:
great. thanks a lot

PR #505, opened 2020-08-12
docs: alert for memeory leak inject to plugins

debs-obrien:


debs-obrien:
Can you make sure this is correct - this is in reference to this pr from old docs site - https://github.com/nuxt/docs/pull/1973

PR #507, opened 2020-08-12
Update installation.md

sambezalel:
'The' was repeated twice on line 154 and I added 'd' to make name on line 176 past tense


debs-obrien:
Thanks for your contribution 😀

PR #512, opened 2020-08-14
docs: add style extension

debs-obrien:


content/en/guides/configuration-glossary/configuration-css.md
```diff
@@ -31,3 +31,21 @@ export default {
...
+
+```js{}[nuxt.config.js]
+export default {
+  css: ['~/assets/css/main', '~/assets/css/animations']
+}
+```
+
+<base-alert>
+
+If you have two files with the same name eg. main.scss and main.css, and don't add an extension in the css array eg. `css:['~/assets/css/main']`, then the files will be loaded in a specific order, first the css file and then the scss file.
```
manniL, line 47: If I remember correctly, only the css file would be loaded in this case.

debs-obrien: ok that's not what you said on discord. I shall test it out and let you know what behaviour is

debs-obrien: ok confirmed. the scss file gets completely ignored. will update. thanks for checking over it while you're on hols

debs-obrien: ok let me know what you think

manniL: > ok that's not what you said on discord. I shall test it out and let you know what behaviour is


---

My words were: 
> In such a case, the app.css file would be used as css is before scss in the extension list.

I'd say that sentence explains the behavior. Could've been more explicit on that and mention that *only* this file could've been used though 😋

content/en/guides/configuration-glossary/configuration-css.md
```diff
@@ -31,3 +31,21 @@ export default {
...
+
+```js{}[nuxt.config.js]
+export default {
+  css: ['~/assets/css/main', '~/assets/css/animations']
+}
+```
+
+<base-alert>
+
+If you have two files with the same name eg. main.scss and main.css, and don't add an extension in the css array eg. `css:['~/assets/css/main']`, then the only one file will be loaded depending on the default order. In this case only the css file will be loaded and the scss file will be ignored.
```
manniL, line 47:
```suggestion
If you have two files with the same name eg. `main.scss` and `main.css`, and don't specify an extension in the css array entry, eg. `css: ['~/assets/css/main']`, then only one file will be loaded depending on the order of `styleExtensions`. In this case only the `css` file will be loaded and the `scss` file will be ignored because `css` comes first in the default `styleExtension` array.
```

Made it a bit clearer, but might be "too verbose"?

debs-obrien: thanks. I think its much better

PR #530, opened 2020-08-20
docs(guides): fix typo on internals-glossary > context guide

oiagorodrigues:
File was showing `\$config` on a heading of the `Context` guide, on `Internal Glossary`.
Maybe it was a scape to something, but it was rendered in the page.

>I've noticed that prettier added this automatically and running the formatter.


debs-obrien:
thanks. yes it was added by the content module to escape but I think this is now fixed so this should work now. Thanks

PR #534, opened 2020-08-21
docs: add information to preview mode

mornir:
Don't hesitate to correct/improve the information 😉


content/en/guides/features/live-preview.md
```diff
@@ -64,17 +64,30 @@ enablePreview should be tested locally with yarn start and not yarn
...
 ```
 
+### Passing data to enablePreview
+
+You can pass data to the `enablePreview` function. That data will then be available on the `$preview` context helper and on `this.$preview`.
+This can be useful if you need to fetch data differently in preview mode (e.g. making authenticated requests to fetch draft content).
+
+```js
+asyncData ({ $preview }) {
+  if ($preview) return { page: $preview }
```
Atinux, line 87: I don't think this example returning `{ page: $preview }` is correct, instead you can fake an api call and give a token from `$preview` instead.

Atinux (changes requested):
Nice work :)

content/en/guides/features/live-preview.md
mornir: You mean you would call `enablePreview({token: 'xxxxxx'})` from the plugin and then access that token inside asyncData? In my example, I assume that the API call with the token is already made inside the plugin and then the data is passed to `enablePreview(data)`.

Atinux: I see, in that case I think we should add more code example for this section to show the plugin and maybe some comments.

Also, I think if you use $axios, you can directly set the token in the preview plugin instead of using `$preview`

mornir: I now wonder if making auth request to preview draft content in headless CMS is a good use case for this feature.
What use cases did you have in mind when adding the `$preview` context helper?

Atinux: Actually I don't have such at the moment 😅

mornir: 😅 In that case it's maybe best to only mention that it's possible to pass data, without giving any example.

Atinux: Yes, can you update it :)

Atinux (approved):

debs-obrien:
@oiagorodrigues can you update Portuguese please with this minor change. thanks 😊

oiagorodrigues:
@debs-obrien sure. I'm on it.

PR #538, opened 2020-08-24
Fix some typo in context.md (ko and en)

seongjin605:
image

As in the picture above, bug can be caused by a markdown syntax problem(korean version).
In addition, please check commit logs for English typo.

Fix them please. 
thank you 😁


debs-obrien:
Thanks, yes we noticed this also in the english version due to TOC and fixed it. Thanks for fixing it in Korean version

seongjin605:
@debs-obrien I am very happy to be able to help nuxt.js users. Thank you very much for reviewing my PR with interest. 😄

PR #558, opened 2020-08-24
Fixed formatting of the trailingSlash section

mattbab:
Fixed formatting of the last two sections in the page. The trailingSlash section appears in the alert from scrollBehavior


debs-obrien:
nice find. thanks 😊

PR #559, opened 2020-08-25
Update data-fetching.md

febrihidayan:
change name to title 😁


debs-obrien:
you're right. nice find 😊

PR #573, opened 2020-08-25
[Indonesia] Improve language structure

febrihidayan:
I did this because looking at the shop title page which is `the NuxtJS Toko` it should be `Toko NuxtJS`

screenshot


debs-obrien:
Thanks @febrihidayan . great work. However I prefer we use i18n interpolation instead of pre and post in lang files. The home page has an example of this so for the shop it would be similar
`<i18n
          path="shop.title"
          tag="h1"
          class=""
        >
          {{ $t('shop.title') }}
          <template v-slot:nuxt>
            <AppTitle />
          </template>
        </i18n>`

Then in the lang file:
id.js
`title: 'Toko {nuxt}',`
en.js
`title: '{nuxt} Shop',`

And add to all other translations.

You will see we started this for the home page which can be found in the components/partials/home and it makes sense to keep this same standard across the site as well as it is just much easier to maintain. Let me know if you need any help with it.

debs-obrien (changes requested):
use i18n interpolation for titles and not pre and post in lang file

febrihidayan:
@debs-obrien well, i will change it as soon as possible

febrihidayan:
@debs-obrien How can this be used in the head()?

debs-obrien:
I would use a meta tag for that like we have done on the index page

febrihidayan:
head
Could you show me the code for this problem

debs-obrien:
meta tags should be used for meta and not the title tag that is used in the head which contains styling etc. The index page has an example of how we have done this
`head() {
    return {
      title: this.$i18n.t('homepage.meta.title'),`
....

`homepage: {
    meta: {
      title: 'Nuxt.js - The Intuitive Vue Framework',
      description:
        'Build your next Vue.js application with confidence using NuxtJS. An open source framework making web development simple and powerful.'
    },`

febrihidayan:
```
...
<i18n path="shop.title" tag="h1" class="">
  {{ $t('shop.title') }}
  <template v-slot:nuxt>
    <AppTitle />
  </template>
</i18n>
...
head() {
  const title = this.$i18n.t('shop.title')

  return {
    title,
    ....
  }
}
...
```

id-ID.js
`shop: { title: 'Toko {nuxt}' }`

do i have to add like this? 🤔
id-ID.js
```
shop: {
  meta: {
    title: 'Toko NuxtJS'
  }
}
```

debs-obrien:
You will have to have both. So meta title and title. The meta will be normal text and the title will have the {nuxt} in it which provides the styling. So like in the index page (homepage in en_EN.js we have a property called meta and then we have titles and other texts. This is how it should be done. Meta should be separate and not include styling

febrihidayan:
Fine, I'll let you know if there are other mistakes 😀

pages/blog/index.vue
```diff
@@ -21,15 +25,15 @@
...
+                {{ $t('blog.nuxt_team') }}
               </NuxtLink>
             </template>
             <template v-slot:ambassadors>
               <NuxtLink
                 class="text-nuxt-green underline"
                 to="/team#nuxtCommunity"
               >
-                Nuxt.js Community
+                {{ $t('blog.ambassadors') }}
```
debs-obrien, line 36: this should be community. community includes ambassadors but it is more than just ambassadors

debs-obrien (changes requested):
looks great. Nice work. just one minor change needed. But really great job 💪

i18n/en-US.js
```diff
@@ -273,9 +273,9 @@ module.exports = {
     },
     title: '{nuxt} Blog',
     description:
-      'Discover articles from the {nuxtTeam} and {ambassadors} about NuxtJS, tips and tricks included!',
+      'Discover articles from the {nuxtTeam} and {nuxtCommunity} about NuxtJS, tips and tricks included!',
     nuxt_team: 'NuxtJS Team',
-    ambassadors: 'NuxtJS Community',
+    nuxt_community: 'NuxtJS Community',
```
debs-obrien, line 278: this needs to be Nuxt.js Community not NuxtJS. NuxtJS is for company and Nuxt.js is for open source. You will see it on the page https://nuxtjs.org/blog
Thanks 😊

debs-obrien (changes requested):
nice, sorry just noticed one minor thing if you could change it that would be awesome

febrihidayan:
I changed it like this 🙄

i18n/en-US.js
febrihidayan: I just found out about it 😅
i will change it back

debs-obrien:
fantastic work. well done.🤩

febrihidayan:
yeee!, finally finished 😊

PR #574, opened 2020-08-25
Fix example

vhoyer:


debs-obrien:
@Atinux can you check this please as I know we created the new example with the nuxt hook inside the addPlugin perhaps to make it simpler so am I right in saying this change is not what we want to show in example code or is it an error on our behalf?

vhoyer:
Yeah, I also think it would be simpler to just have the nuxt hook there, as it makes the reader have to assume less things to understand the code.

I added the rest of the addPlugin() bit, because, as I saw, the syntax wasn't correct. But I may be wrong there

If it helps the discussion, I think it's best for this example to only have the `nuxt.hook` line and remove the `addPlugin`.

(I also fixed the identation, there were tabs mixed with spaces, so the identation was uneven :smile:)

Atinux:
Thank you @vhoyer 

Indeed we had a syntax error, I don't understad why we have `this.addPlugin(` @debs-obrien, I guess it is a typo?

For this example, I think we can remove the `addPlugin()` block to keep it simple :)

debs-obrien:
Thanks @vhoyer 
@Atinux I think it was just copied over from our old docs example and somehow got mixed into this example. 🤷‍♀️

PR #578, opened 2020-08-26
[Spanish] translate new guides

rsdiaz:
I have edited to correct some mistakes in the translation.


debs-obrien:
fantastic. Thanks 😊

PR #585, opened 2020-08-27
Translating content/es/guides/get-started/conclusion.md into Spanish

alvaroOnCode:
I hope it works!

### DIFF
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -197,7 +197,8 @@ export default {
         file: 'pt-BR.js',
         name: 'Português',
         domain: 'https://pt.nuxtjs.org'
-      },
+      }
+      // please leave this commented out until subdomain is live
       // {
       // code: 'tr',
       // iso: 'tr-TR',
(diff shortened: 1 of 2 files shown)


nuxt.config.js
```diff
@@ -197,21 +197,21 @@ export default {
...
       // domain: 'https://tr.nuxtjs.org'
       // }
-      // {
-      //   code: 'es',
-      //   iso: 'es-ES',
-      //   file: 'es-ES.js',
-      //   name: 'Español',
-      //   domain: 'https://es.nuxtjs.org'
-      // }
+      {
```
debs-obrien, line 208: we need to leave this commented out for now until the domain is ready otherwise it will appear in the footer select with a broken link.

debs-obrien:
hi @alvaroOnCode page looks great just the nuxt config needs not to be changed for nows. if you could fix that it would be great and make sure you tick the page off in the issues so people can know what has been done. Thanks for your contribution.

nuxt.config.js
debs-obrien: i sorted it and added a comment for future contributors

alvaroOnCode: Great, thank you!

Issue #543, opened 2020-08-24
[Spanish] translate new guides


debs-obrien:
Looking for help to translate our website to Spanish. It has already been setup but we would love some help translating the guides folder. See the readMe for info on how to translate our site: https://github.com/nuxt/nuxtjs.org

Feel free to just take one page and submit a pr. make sure you also tick the box in the list below. Or take on the whole project and just create a draft pr and we will merge it when it is finished.

i18n configuration

- [x]  i18n file
- [x]  i18n block on nuxt.config

📝 guides -> components glossary

- [x]  pages-fetch.md
- [x]  pages-head.md
- [x]  pages-key.md
- [x]  pages-layout.md
- [x]  pages-loading.md
- [x]  pages-middleware.md
- [x]  pages-scrolltotop.md
- [x]  pages-transition.md
- [x]  pages-validate.md
- [x]  pages-watchquery.md

📝 guides -> concepts

- [x]  context-helpers.md
- [x]  nuxt-lifecycle.md
- [x]  server-side-rendering.md
- [x]  static-site-generation.md
- [x]  views.md

📝 guides -> configuration glossary
 

- [x] configuration-build.md
- [x]  configuration-generate.md
- [x]  configuration-render.md
- [x]  configuration-server.md
- [x]  configuration-telemetry.md
- [x]  configuration-router.md
- [x]  configuration-server-middleware.md
- [x]  configuration-transition.md
- [x]  configuration-builddir.md
- [x]  configuration-cli.md
- [x]  configuration-components.md
- [x]  configuration-css.md
- [x]  configuration-dev.md
- [x]  configuration-dir.md
- [x]  configuration-env.md
- [x]  configuration-extend-plugins.md
- [x]  configuration-global-name.md
- [x]  configuration-head.md
- [x]  configuration-hooks.md
- [x]  configuration-ignore.md
- [x]  configuration-loading.md
- [x]  configuration-loading-indicator.md
- [x]  configuration-mode.md
- [x]  configuration-modern.md
- [x]  configuration-modules.md
- [x]  configuration-modulesdir.md
- [x]  configuration-plugins.md
- [x]  configuration-rootdir.md
- [x]  configuration-runtime-config.md
- [x]  configuration-srcdir.md
- [x]  configuration-target.md
- [x]  configuration-vue-config.md
- [x]  configuration-watch.md
- [x]  configuration-watchers.md

📝 guides -> directory structure

- [x]  assets.md
- [x]  components.md
[...]

PR #598, opened 2020-08-27
translate to spanish guides -> components-glossary

mblayer:
This PR is a WIP, Includes the features:

📝 guides -> components glossary

- [x] pages-loading.md
- [x] pages-middleware.md
- [x] pages-scrolltotop.md
- [x] pages-transition.md
- [x] pages-validate.md


mblayer:
> Hi! The usage of formal Spanish in this file is fine but in the rest of your commits you are using informal.
> I would suggest to change it back to informal, as most of the translators agreed upon it, as you can read here: #543

I had translated it before the agreement, there I see to update it, thanks for the advice

debs-obrien:
great work, thanks 😃

PR #606, opened 2020-08-28
[korean] Translate file-system-routing.md

seongjin605:
Also some markdown files fixed
thank you.


debs-obrien:
Thanks @seongjin605 can you make sure you only submit Korean with this pr as mixing languages is hard to review especially as others are working on those lang files. If there is a fix you find in another lang file then please go ahead and submit it but as a separate pr. Thanks

PR #608, opened 2020-08-28
Persian i18n added and translated

AdnanBabakan:


nuxt.config.js
```diff
@@ -197,6 +201,13 @@ export default {
         file: 'pt-BR.js',
         name: 'Português',
         domain: 'https://pt.nuxtjs.org'
+      },
```
debs-obrien, line 204: this lang cannot be put live yet until the subdomain has been created so for now you can leave it commented out.

nuxt.config.js
```diff
@@ -2,6 +2,10 @@ export default {
   target: 'static',
   ssr: true,
   head: {
+    htmlAttrs: {
+      lang: 'fa',
```
debs-obrien, line 6: this will be added to every single language if I am not mistaken so we need to add rtl only if the lang is fa.

debs-obrien (commented):
Great start and nice adding the rtl. this is the first rtl lang we have for this site. Just a few comments I left regarding the nuxt config. To make it easier for others to collaborate I would just copy over the file you have translated and that means others in the issue can see which files/chapters are missing rather than having to open each md file to see if it needs translating. But just my opinion. It is what the others have been doing that are working on other languages and seems to be working quite well.

nuxt.config.js
AdnanBabakan: Sure

nuxt.config.js
AdnanBabakan: So how do you suggest to do it? What's the proper way to do so? Is there a way to get the current language and act according to that?

debs-obrien: good question. we will have to try something using if i18n.locale === fa add rtl. Never tried it before but give it a go and if you get stuck just comment it out and we will look into it next week.

AdnanBabakan: I tried this and it is working perfectly:

```javascript
    htmlAttrs: {
      lang: process.env.NUXT_LOCALE,
      dir: ['fa', 'ar', 'he'].includes(process.env.NUXT_LOCALE) ? 'rtl' : 'ltr'
    }
```

And for the production we can for sure generate the website using the local wanted.

The only three languages I know which are RTL are Persian, Arabic and Hebrew so I added them to the array.

debs-obrien: fantastic. great job 💪

debs-obrien:
so @AdnanBabakan  what do you think of removing the non translated files to make it easier for others to just copy one over and work on it rather than having them all there

AdnanBabakan:
> so @AdnanBabakan what do you think of removing the non translated files to make it easier for others to just copy one over and work on it rather than having them all there

Of course. I already did it after I read your comment, the new commit will have the translated files only.

debs-obrien:
Fantastic some great work done there. Well done. Just make sure the nuxt config has the lang commented out like the spanish language. By adding it here it adds it automatically to the footer select so we can't show it till domain is live. 
Other than that everything else looks great

AdnanBabakan:
Sure

PR #639, opened 2020-09-02
Fix Nuxtjs.org page broken layout

Donghoon759:
## When I open the Nuxtjs.org page on Chrome, it looks like this and I figured out the reason. ##
image

## ```svg``` class is added globally so that the width is limited to 20px. ##
image

Styles in ```AppModal.vue``` and ```BaseAlert.vue``` are added globally. I think there is no reason they have to be added globally because **the classes in those files are not used in other components at all.**
### So I think it is better to limit their styles to be scoped and it will fix the broken layout issue. ###
Thanks.

### DIFF
--- a/components/global/bases/AppModal.vue
+++ b/components/global/bases/AppModal.vue
@@ -81,7 +81,7 @@ export default {
 }
 </script>
 
-<style>
+<style scoped>
 .svg {
   width: 20px;
   color: black;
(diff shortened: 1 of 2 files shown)


debs-obrien:
interesting. I can't reproduce this error. for me it shows the page working perfect. but I do agree that the styles should be scoped. Thanks for investigating it and let me know after merge if it is ok. thanks

PR #650, opened 2020-09-03
refactor(i18n): add footer translations to pt-br

oiagorodrigues:
Translated the Footer links to pt-BR.

image


debs-obrien:
Nice 😃

Issue #468, opened 2020-08-05
Change the video url on the homepage


febrihidayan:
I get bad news where the url https://player.vimeo.com/video/311756540 with the vimeo.com domain has been blocked.
This blocking applies in the country of Indonesia, it is blocked based on expressions from [Kominfo Indonesia for vimeo.com](https://kominfo.go.id/index.php/content/detail/3984/Siaran+Pers+No.+30+-PIH-Kominfo-5-2014+tentang++PENANGANAN+VIMEO.COM+/0/siaran_pers). So, this blocking makes the video blank.

I hope YouTube is much better, [Vue Mastery video](https://www.youtube.com/embed/7ITypVi-qRY)

Thanks

PR #661, opened 2020-09-05
fix: video not showing for Indonesia

febrihidayan:
this happens because the vimeo.com domain is blocked in Indonesia #468 

video not showing for Indonesia

### DIFF
--- a/components/partials/home/HomeWelcome.vue
+++ b/components/partials/home/HomeWelcome.vue
@@ -122,14 +122,11 @@ export default {
     MeteorIcon,
     GithubIcon
   },
-  data() {
-    return {
-      videoUrl: 'https://player.vimeo.com/video/311756540'
-    }
-  },
-  mounted() {
-    if (this.$i18n.locale === 'id') {
-      this.videoUrl = 'https://www.youtube.com/embed/7ITypVi-qRY'
+  computed: {
+    videoUrl() {
+      return this.$i18n.locale === 'id'
+        ? 'https://www.youtube.com/embed/7ITypVi-qRY'
+        : 'https://player.vimeo.com/video/311756540'
     }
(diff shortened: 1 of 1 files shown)


manniL:
This could be implemented as computed property instead of checking solely on mounted.

manniL (approved):

components/partials/home/HomeWelcome.vue
```diff
@@ -121,6 +121,13 @@ export default {
   components: {
     MeteorIcon,
     GithubIcon
+  },
+  computed: {
+    videoUrl() {
+      return this.$i18n.locale === 'id'
```
manniL, line 127: nit: Ideally leave a comment here linking to this PR (or alternatively describing why the video URL is different for the indonesian language.

febrihidayan: I did it so that the video display appeared in Indonesia. because the vimeo.com site is prohibited or blocked in Indonesia. So it's no wonder I do this so that the video continues to appear on nuxtjs.org by using the youtube service for the video

manniL: Yes, I'm aware of that. But I think having at least a reference to that PR (or a short explanation) *in the code* would help future readers to understand why the code has been changed like that :)

febrihidayan: I've fixed it, you mean like that

debs-obrien:
great job @febrihidayan I would have never known that vimeo is blocked in Indonesia. Interesting

components/partials/home/HomeWelcome.vue
manniL: Yes, perfect 👍🏻

PR #664, opened 2020-09-05
docs: Add deploy to Bip tutorial

joelkennedy:
## Description

Added a short tutorial on how to deploy to [Bip](https://bip.sh). Bip contains automatic framework detection for NuxtJS, so this is reflected in the guide. 

An example NuxtJS project hosted on Bip is available at https://nuxtjsexample.bip.sh, which utilises [create-nuxt-app](https://github.com/nuxt/create-nuxt-app).


debs-obrien:
great job. FYI for next time you can add a category number with a point which saves you having to modify all other categories. so 206.1 would work and would be between 206 and 207 for example. Thanks for the contribution 😃

PR #670, opened 2020-09-07
docs: improve rendering and depolyment pages

debs-obrien:


Atinux (approved):

content/en/guides/features/deployment-targets.md
```diff
@@ -7,31 +7,62 @@ category: features
 
 ## Static Hosting
 
-For static hosting (hosting where no server is needed) the target of static needs to be added to your nuxt.config file.
+With static hosting, hosting where no server is needed, you can choose to either host a single page application or a multiple page application, also known as static sties. With static hosting no server is needed, meaning your SPA or static site can be hosted on any serverless hosting or CDN for free. For static hosting the target of `static` needs to be added to your `nuxt.config` file.
```
pi0, line 10: I think would be better to use `any static hosting` instead of `any serverless hosting` because serverless target is server  + not always is free.

content/en/guides/features/deployment-targets.md
```diff
@@ -7,31 +7,62 @@ category: features
...
 ```
 
-Running nuxt dev with the static target will improve the developer experience:
+### SPA
+
+Single page applications are pages that are rendered only on the client side without the need of a server. To deploy a single page application set the [mode set to `spa`](/guides/features/rendering-modes#spa) and then use the `build` command to build your application.
+
+```js{}[nuxt.config.js]
+export default {
+  target: 'static' // default is 'server'
```
pi0, line 24: Command be `nuxt generate` but i don't think we should set target to either `static` or `server`. SPA with `server` target is also valid so users don't need to specify it.

content/en/guides/features/deployment-targets.md
```diff
@@ -7,31 +7,62 @@ category: features
...
+  mode: 'spa'
+}
+```
+
+### Static Sites
+
+As Nuxt.js also works as a static site generator you can therefore generate your application as a static site. Statically render your Nuxt.js application and get all of benefits of a universal app without a server. The nuxt `generate` command will generate a completely static version of your website. It will generate HTML for every one of your routes and put it inside of its own file in the `dist` folder. Basically any .vue file that is placed inside the pages folder will be generated as a static html page. This improves performance as well as SEO and better offline support.
```
pi0, line 31: > Basically any .vue file that is placed inside the pages folder will be generated as a static html page

Is mentioning `.vue` necessary? Because it can be also other formats. We could reword to any page file you create for example.

content/en/guides/features/deployment-targets.md
```diff
@@ -7,31 +7,62 @@ category: features
...
 <base-alert type="info">
+
 We are also exposing `process.target` for module authors to add logic depending on the user target.
+
 </base-alert>
 
 ## Server Hosting
 
-For server hosting the target of server is used which is the default value.
+Server Hosting is hosting that requires a server and is intended for SSR applications. Server-side rendering otherwise known as SSR means that your page is rendered on the server when it is requested by the user. When the user opens your page in a browser the browser sends a request to the server requesting that page. The page is rendered on the server and sent back to the browser with all its content.
```
pi0, line 54: Here we are describing `Server Side Rendering` or `Isomorphic Rendering` not hosting! I think we should rename it to `Server Rendering`. Server Hosting, as mentioned above is also valid for SPA (with API) or serverless functions.

content/en/guides/features/deployment-targets.md
debs-obrien: this is where there is a lot of confusion. yes SPA can be either on a static hosting or a server hosting. SPA on netlify or azure static web apps so the deployment target should be static as it is a static deployment (for the sake of logic). So if you don't use target static or server then we need to say for single page applications there is no need to set the target or something similar and for spa you only need to modify the mode. It is as if these 2 pages should almost be one page that shows the 3 options with their corresponding targets and modes. Any suggestions?

content/en/guides/features/deployment-targets.md
debs-obrien: suggestion: https://github.com/nuxt/nuxtjs.org/pull/675/files

PR #679, opened 2020-09-08
[japanese] add company on using NuxtJS section

k-miyake:
@Atinux
The new service I've launched in my company was developed in Nuxt.js, 
and I'd like to put my logo on "using NuxtJS" section of the Japanese site.
I asked @kazupon about this.


k-miyake:
@Atinux Thanks!

PR #675, opened 2020-09-07
Fix target mode docs

debs-obrien:

PR #689, opened 2020-09-11
Fixed typo & grammar

CaddyDz:
Not sure about the LF


debs-obrien:
Thanks, we have just improved this page with this pr: https://github.com/nuxt/nuxtjs.org/pull/675

PR #690, opened 2020-09-12
Improve data fetching hook documentation

thomasboyt:
I've been migrating my app to using the new fetch hooks, but I thought the documentation was rather confusing. I'm hoping that the changes I've made here will make them easier to understand:

* Better-describe differences between `fetch` and `asyncData`, particularly the distinction between fetch's "render with placeholder" and asyncData's "block navigation" strategies on the client

* Remove `@nuxt/http` setup guide which is covered by the linked documentation (if it belongs in this documentation, it probably should get its own section, since it could be used by `fetch` just as well)


alexchopin:
Hi @thomasboyt,
Thanks for your great contribution, it's a really nice improvement.

Issue #550, opened 2020-08-24
[French] translate new guides


debs-obrien:
French translation is almost finished under this draft pr:https://github.com/nuxt/nuxtjs.org/pull/704

kissu:
I've created a draft PR to translate the whole 🇫🇷 guides at #704.
Do you want me to add you as a maintainer @antoineneff ? 😄 
I may ask my GF @spookyumi to review the translations later on for some advanced grammar (she was a speech therapist not so long ago).

lihbr:
Hey @kissu if you can invite me to your fork that'd be awesome! I'll try to free up some time to help there 🙂
Happy to make some commits there if we're going one big PR for the French translation~

kissu:
I've added you and @roiLeo on my PR. 😄 
Thanks for helping ! 🙏🏻

kissu:
Btw folks, I want to be sure (or agree on a point of view), do you translate some of those:
- Code splitting into _fragmentation du code_
- Lifecycle into _cycle de vie_
- Server Side Rendering (SSR) into _rendu côté serveur (SSR)_

I'm asking that because as far as I know, people do not translate those but keep it in English, for future googling or even for ease of use when talking about those principles in an English place. Even myself, I do only use those terms.

And, documentations usually tend to keep those in English as far as I know (even if I do not read that much of :fr: documentation anymore 🙄).
So, do we agree to translate those specific words or keep them in English? I'd rather keep those in English. 😁

lihbr:
That's a nice question 🤔 I tend to think that the point of a translated documentation is that it's in part addressed to people struggling with English so perhaps it's best to be as inclusive as possible by also translating those terms.

Does _Rendu Côté Serveur_ has to become _RCS_ then? _Générateur de Site Statique_ > _GSS_? 🤔

Perhaps @debs-obrien you have some "official" guidance there from the almost translated Spanish doc ☺️

kissu:
She may take a bit of time answering because of [that](https://twitter.com/debs_obrien/status/1308870712787116032). She hopefully feels better today ! 🎉

debs-obrien:
[...]

PR #698, opened 2020-09-14
docs(fr): fix guides/get-started/routing

antoineneff:
Fix french translation (typos, grammar and some rewrites) for guides/get-started/routing page.

I will continue to fix translations and might begin to translate missing pages, do you want me to have a complete directory to PR instead of each page ?


debs-obrien:
hey great thanks. whichever works. just add a comment in this pr: https://github.com/nuxt/nuxtjs.org/issues/550 that way its easier for others to track what you are doing

kissu:
I'm currently starting doing the whole 🇫🇷 translations.
We may pair/split our work on each sections if you wish @antoineneff. 😉

antoineneff:
@kissu Hey sure, just share what you're doing in the issue #550 so we don't work on the same pages.
And if you want, we can review each other translations, just mention me.

PR #702, opened 2020-09-15
fix: typo

manniL:


debs-obrien:
why do my fingers not type as fast as my mind

PR #708, opened 2020-09-15
Translation to Farsi(Persian) - upgrading.md

MahdiFakhr:
Translation of the following document guides/get-started/upgrading.md

**note:** Some words like "static hosting" do not have good equivalents in Persian. Due to this reason, it is better to use the same English word in Persian to prevent confusion.


debs-obrien:
Great, thanks

PR #711, opened 2020-09-16
Added example of how to use serverMiddleware as an API endpoint

adelriosantiago:
From SO question: https://stackoverflow.com/questions/62202526/how-do-i-get-the-post-data-from-a-nuxtjs-server-middleware


Atinux (approved):

manniL:
Would it make sense to add the npm/yarn command for the deps used (express & body-parser) so the guide is "complete"? 😋

debs-obrien:
Thanks guys. @manniL feel free to make a pr for any additions you consider worthwhile.

PR #713, opened 2020-09-16
Update going-full-static.md

makma:
Add note that full static is not available when using mode: 'spa' or ssr: 'false'


manniL:
I'd suggest to rewrite it to `ssr: true`/`ssr: false` and provide the legacy `mode` parameter only as a hint.

debs-obrien:
`ssr:true` can actually be confusing especially when mentioning static sites. Even though `ssr:true` is default method we don't encourage adding it to your nuxt.config and therefore just encourage adding `ssr:false` for spa mode. I do agree that we need to better reword this and perhaps as @manniL said provide the legacy mode only as a hint with a note that it is deprecated.

makma:
I'm a little bit confused again 😵. Do I understand correctly that `mode` property has been already deprecated with **both** `spa` and `universal` values?
And the new mappings are `mode: 'spa'` -> `ssr: 'false'` and `mode: 'universal'` -> `ssr: 'true'`.
Am I correct? Thank you.

manniL:
> I'm a little bit confused again 😵. Do I understand correctly that `mode` property has been already deprecated with **both** `spa` and `universal` values?
> And the new mappings are `mode: 'spa'` -> `ssr: 'false'` and `mode: 'universal'` -> `ssr: 'true'`.
> Am I correct? Thank you.

Both correct, yes. It has been deprecated &  will be replaced. "ssr: true/false" should be easier to grasp for people.

makma:
> > I'm a little bit confused again 😵. Do I understand correctly that `mode` property has been already deprecated with **both** `spa` and `universal` values?
> > And the new mappings are `mode: 'spa'` -> `ssr: 'false'` and `mode: 'universal'` -> `ssr: 'true'`.
> > Am I correct? Thank you.
> 
> Both correct, yes. It has been deprecated & will be replaced. "ssr: true/false" should be easier to grasp for people.

Thank you, I've rephrased it in a new commit.

debs-obrien:
Thanks @makma 😃

manniL:
Short addendum (related to https://twitter.com/ivanprats94/status/1311354798143348738).

You **can** server your SPA (`ssr: false`) built with Nuxt through a CDN. But using the new "full static mode" is not possible when SSR is disabled.

PR #716, opened 2020-09-17
Update routing.md

katerlouis:
Vue suggests to you kebab-case for components.
The docs already do this here: https://nuxtjs.org/guides/features/file-system-routing and should be consistent in this.

### DIFF
--- a/content/en/guides/get-started/routing.md
+++ b/content/en/guides/get-started/routing.md
@@ -20,25 +20,25 @@ Learn more about [Routing](/guides/features/file-system-routing)
 
 ## Navigation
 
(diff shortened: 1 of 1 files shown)


manniL:
> Vue suggests to you kebab-case for components.

Yes, but only when using DOM templates. In SFCs, `PascalCase` is recommended ([source](https://vuejs.org/v2/style-guide/#Component-name-casing-in-templates-strongly-recommended)). Thus, we decided to use `PascalCase` where appropriate. If there are inconsistencies, please feel free to send in a PR for them there.

manniL (changes requested):
See comment above

debs-obrien:
Hi @katerlouis thanks for your contribution. As @manniL stated PascalCase is recommenced for SFC and it is what we have decided to use as a standard thought our docs and code examples and site. It doesn't mean that you can't use kebab-case just that we are being consistent and only recommending one way of writing it to make things easier for people and will therefore recommend to use <NuxtLink> over <nuxt-link>. Hope that helps explain why it is this way in the docs. Reach out if you need further clarification.

katerlouis:
Agreed. Created another PR for the different usage in the docs. :)

PR #721, opened 2020-09-17
Fix structure of layouts.default.vue and path

davidoskay:
"Fix error: "Component template should contain exactly one root element.."


debs-obrien:
nice catch, thanks

PR #727, opened 2020-09-19
Changed menu item of SSR config item

bcnzer:
In the documentation the Configuration glossary has two "mode" properties. The second, further down in the list, was actually SSR. It was alphabetically placed where SSR should be.

This change simply changes the menu item. 

Background: I couldn't find the documentation on the SSR property of the nuxt.config file. Discovered it by accident by going to Mode and seeing a link to SSR, which linked to the second "mode" entry.


debs-obrien:
oops, thanks @bcnzer

PR #735, opened 2020-09-19
fix:  svg of 404 page is responsive now

ali-furkqn:
While browsing the nuxtjs.org, I noticed that the 404 page was mistake ( LostImageIllustration's position).


debs-obrien:
Great, thanks for your contribution

PR #756, opened 2020-09-22
feat: add examples book

debs-obrien:


Kronhyx (approved):

debs-obrien:
@Atinux this is the examples book which with descriptions and button to open code sandbox within the other books. Live demo can be seen here: 
https://feat-codesandboxes-description--nuxt.netlify.app/guides/examples/hello-world

We can modify intro text or see example button if you prefer. 

Descriptions are in two places so we have to maintain them in both the examples book and in the guides books under the corresponding chapters

All feedback welcome @pi0 @manniL @danielroe

Atinux (dismissed):
Amazing work!

content/en/guides/directory-structure/nuxt-config.md
```diff
@@ -327,6 +327,23 @@ See more on the [transition property](/guides/configuration-glossary/configura
 
 </base-alert>
 
+### Example
+
+<example-intro></example-intro>
+
+`nuxt.config.js` contains :
+
+- the `env` property to add the URL of our API.
```
Atinux, line 336: Could we use the `publicRuntimeConfig` instead of `env` for this example?

debs-obrien: you mean also change it in the docs: we only document env property and not publicRuntimeConfig / privateRuntimeConfig

https://nuxtjs.org/guides/directory-structure/nuxt-config#env

do we remove env from docs altogether or just add the extra ones to this page and use in example? @Atinux

debs-obrien: done

debs-obrien:
runtimeConfig added to docs as well as example: https://github.com/nuxt/nuxtjs.org/pull/890

Atinux:
I think we should use publicRuntimeConfig.axios.baseURL, see https://axios.nuxtjs.org/options

+
```diff
@@ -0,0 +1,6 @@
+Merge branch 'master' into feat-codesandboxes-description
```
pi0, line 1: Accidental add?

i18n/fa-IR.js
```diff
@@ -291,6 +291,9 @@ module.exports = {
     msg1: 'مشارکت در این صفحه بسته شده است. اگر می‌‌خواهید مشارکت کنید',
     link: 'مستندات جدید'
   },
+  example: {
+    intro: 'In this example:'
```
pi0, line 295:
```suggestion
    intro: 'در این مثال:'
```

+
debs-obrien: why? there were some modifications to the docs so I needed to fix merge issues locally so merged master into the branch.

pi0: Actually it is a file named `+`  and staged 😁 

https://github.com/nuxt/nuxtjs.org/blob/a8b4d84b3c01731a6da064e731468da1616a6b15/%2B

debs-obrien: wow. how did that get there. very weird. Thanks. Removed

i18n/fa-IR.js
febrihidayan: @pi0 I'm not sure it's written that way. should be like this `:در این مثال`

PR #758, opened 2020-09-22
refactor: move icons to a reusable icon component

azrikahar:
## Context

I have noticed the icons or specifically SVG assets are imported using the syntax as suggested from the nuxt svg module [over here](https://github.com/nuxt-community/svg-module#vue-svg-loader), which is the following format (as example):

```vue
<template>
  <NuxtLogo />
</template>

<script>
  import NuxtLogo from "~/assets/nuxt.svg?inline";

  export default {
    components: { NuxtLogo },
  };
</script>
```

However when I was using this approach, it started to feel redundant or sort of repetitive, especially in instances like wanting to display a Github icon in 2 pages respectively. This means we'll need to have 1 line for import and 3 lines for `components` option in the export default just to use the same exact icon.

## Solution

This is the reason why I created a component that will load the icon we want based on the prop `name` in my personal projects, and I think it's useful to introduce it here as well. There is a also validator in place to assure that the name we used actually exists in the list of available/defined icons. This makes us able to use the `<AppIcon />` component freely as a single line of code, remove a lot of redundant import statements and even remove the need for `components` option in many cases!

I think another added benefit of this approach is we can just check the AppIcon component to see what are the icons available without having to go through the assets folder manually. Only when we need to import a new icon then we'll just need to do it once.

In case anyone wonders why we can't use the dynamic require approach such as `require('~/assets/icons/${name}.svg?inline')` instead and call it a day, it is because webpack doesn't know what value will we pass for `name` variable, thus it will just include **every** assets inside `~/assets/icons` instead. Reference: https://webpack.js.org/guides/dependency-management/#require-with-expression

Notable quotes from referenced Webpack dynamic require docs:

> A context module is generated. It contains references to **all modules in that directory** that can be required with a request matching the regular expression.

> This means dynamic requires are supported but will cause all matching modules to be included in the bundle.

As there might be assets in the icons folder that are just being put there without being use, I opt for the tree-shakable approach instead. With that said, I think the part where this solution shines the most is in this scenario:
https://github.com/nuxt/nuxtjs.org/blob/53dfa576d0408568d2114ff9e7c618a5f1d59d48/components/TheMobileBottomNav.vue#L15-L21

Here the `link.slug` will pass us the icon name for the mobile bottom nav icon, then it is dynamically determined with the `component is:` solution while appending `-icon` to the string since the imported icon components has `Icon` at the end, like so:
https://github.com/nuxt/nuxtjs.org/blob/53dfa576d0408568d2114ff9e7c618a5f1d59d48/components/TheMobileBottomNav.vue#L60-L65

In this instance, the `<AppIcon />` component is already able to simplify this just by passing the `link.slug` to the `name` prop, so no string concatenation is needed.

---

### Side note 1

I was using `<AppIcon />` as a **functional** component in my projects (for the perfomance gain in Vue 2), but since the performance gain is negligible in Vue 3 , the `functional` attribute is  going to be deprecated. Reference: https://v3.vuejs.org/guide/migration/functional-components.html#overview

And since Nuxt 3 is on the horizon, and I'm aware this site is under an overhaul development which I promptly assume should be using Nuxt 3, I opt to make it as a normal component in this PR.

### Side note 2

If this approach is acceptable, I will make additional PRs for Illustrations as well (`~/assets/illustrations` imported components). So do let me know if you guys think I can proceed with the additional PRs or just halt the PRs if this approach is already being or going to be implemented in the next overhaul.

### DIFF
--- a/components/TheAsideNavNewDocs.vue
+++ b/components/TheAsideNavNewDocs.vue
@@ -24,11 +24,12 @@
               'font-bold': $route.params.book === group
             }"
           >
-            <ChevronDownIcon
+            <AppIcon
               v-if="$route.params.book === group"
+              name="chevronDown"
               class="w-4 h-4 mr-2"
             />
-            <ChevronRightIcon v-else class="w-4 h-4 mr-2" />
+            <AppIcon v-else name="chevronRight" class="w-4 h-4 mr-2" />
             <span>{{ $t(`content.guides.${group}`) }}</span>
           </component>
           <ul v-if="$route.params.book === group" class="pb-8 pl-2">
@@ -70,14 +71,8 @@
 
 <script>
 import sortBy from 'lodash.sortby'
-import ChevronDownIcon from '~/assets/icons/chevron-down.svg?inline'
-import ChevronRightIcon from '~/assets/icons/chevron-right.svg?inline'
 
 export default {
-  components: {
-    ChevronDownIcon,
-    ChevronRightIcon
-  },
   props: {
     links: {
       type: Object,
--- a/components/TheFooter.vue
+++ b/components/TheFooter.vue
@@ -77,7 +77,7 @@
               label="Select language"
             >
               <template v-slot:icon>
-                <GlobeIcon />
+                <AppIcon name="globe" />
               </template>
             </AppLangSelect>
           </div>
@@ -88,12 +88,10 @@
 </template>
 
 <script>
-import GlobeIcon from '~/assets/icons/globe.svg?inline'
 import SmallNuxtLogo from '~/assets/images/logo-small.svg?inline'
 
 export default {
   components: {
-    GlobeIcon,
     SmallNuxtLogo
   },
   data() {
--- a/components/TheMobileAsideNav.vue
+++ b/components/TheMobileAsideNav.vue
@@ -39,7 +39,8 @@
           class="inner-button sm:hidden absolute h-10 w-10 flex items-center justify-center text-nuxt-gray bg-gray-200 dark:bg-dark-elevatedSurface dark:text-dark-onSurfaceSecondary transition-colors duration-300 ease-linear"
           @click="show = false"
         >
-          <TimesIcon
+          <AppIcon
+            name="times"
             class="block h-5 fill-current transition-colors duration-300 ease-linear"
           />
         </button>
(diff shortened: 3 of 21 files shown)


manniL:
But aren't we loading all the icons on a page where we might only need one of them with the approach in your PR? 🤔 

I personally would keep the icons as separate components but wokr on removing import + declaration, based on the component auto import (https://github.com/nuxt/components), instead.

azrikahar:
> But aren't we loading all the icons on a page where we might only need one of them with the approach in your PR? 🤔

Didn't thought about it this way 😨 definitely is loading all icons. Out of curiosity, do you think changing `require()` to function import which is `() => import()` can make this single component work? We can make them into multiple components for Nuxt Components, but I'm just hoping to explore a way to consolidate them haha.

I did do some investigation and here are my findings (filtered to js file size only):

Note: I went from index page, then the `double-icons` page and lastly the `all-icons` page in that order.

###  Direct Import

| Direct Import | One Icon | Two Icons | All Icons |
|---|---|---|---|
| screenshot | chrome_zafhaKI8EP | chrome_tEmaFJlf3n | chrome_5iBWbiPEEM |
| filesize (kb) | 80.8 | 80.8 | 80.8 |

This shows that the icons are bundled with the pages one by one, but since Nuxt has prefetch for links, we loaded all 3 pages to 80.8kb from the first page.

### Using component, but with Require()

| Require() | One Icon | Two Icons | All Icons |
|---|---|---|---|
| screenshot | chrome_YVdpVQIhUx | chrome_CTH7E8wpDj | chrome_Ehu5mfJxxm |
| gzipped filesize (kb) | 79.7 | 79.7 | 79.7 |

This shows that the icons are bundled with the pages one by one, but since Nuxt has prefetch for links, we loaded all 3 pages to 80.8kb from the first page.
### Using component, but with lazy import


| Require() | One Icon | Two Icons | All Icons |
|---|---|---|---|
| screenshot | chrome_ADd3egtRO3 | chrome_v0MLx2C1Zt | chrome_pz4xtFbuSG |
| gzipped filesize (kb) | 79.5 | 80.8 | 92.6 |

debs-obrien:
Really great work and impressive explanations and research. You are right in saying that we are working on a complete new version of the website so many things will be redundant but it is still good to see things from different points of view and perhaps then add them here or to the new site so all your work and efforts are very much appreciated. I will as @alexchopin to review this and give you his opinion on if you should proceed with the pr as he is the architect on this project. Thanks again for your great work

debs-obrien:
Thanks for your amazing work. We are working on a new design for our website right now and everything is being updated. So for now I will close this pr and soon you will see major changes throughout the website. Thanks

PR #764, opened 2020-09-24
Fixed mistakes I found

GithubProfile:


GithubProfile:
To elaborate more on "simple/example naming inconsistency":
In content/en/guides/directory-structure/modules.md there's a Write your own Module section. It has two code blocks which go one after another. In first block module is called "simple". But next block, which, as I understand, directly corresponds to the first one - shows code from "modules/example.js". You can view it here - https://nuxtjs.org/guides/directory-structure/modules#write-your-own-module. It seems [modules/example.js] doesn't show up if you view modules.md on GitHub.

debs-obrien:
Nice @GithubProfile . thanks for the great contribution and finding those little errors. Appreciate it 💪

GithubProfile:
Thanks.

PR #781, opened 2020-09-30
improved Docs

FalakSarai:


debs-obrien:
Thanks for your contribution but I prefer to not add this line. Being the best for sure in our opinion but we can not officially state it.

PR #784, opened 2020-10-01
Update Development doc.

MAYANK25402:


debs-obrien:
Thanks for your contribution but I think it is best to leave it as just Development as easier when quickly scanning the page and it is a familiar word so it works good as a title, just like how below it is Production.

PR #787, opened 2020-10-01
fix: clarify Vuex store activation

schweikart:
This pull request clarifies the activation of the Vuex store using the `store` directory. The guides previously stated that creating a `store` directory is sufficient to activate the store but the store is in fact only activated if the directory exists and contains a non-hidden file that is not `README.md` (this can be seen in the [`options.js`](https://github.com/nuxt/nuxt.js/blob/dev/packages/config/src/options.js#L228) code).

The less precise variant is usually enough because in most cases the store is activated to create custom store modules. In some cases however, the store is only used by module dependencies like `@nuxtjs/auth` and must be activated without the need for a custom store js-file.

Both the old `/guide` and the new `/guides` pages on this topic were updated.


debs-obrien:
Hi, I think you are right in that it is missing further explanation but I think we could make it sound a little easier to understand:
If it contains a file, that isn't a hidden file or a ReadMe.md file, then the store will be activated.
Or something similar to that.

schweikart:
Yeah I see what you mean. I'd suggest the following:
```md
Nuxt.js will look for the `store` directory. If it contains a file, that isn't a hidden file or a `README.md` file, then the store will be activated. This means that Nuxt will:

1. Import Vuex,
2. Add the `store` option to the root Vue instance.
```
It's a bit long but it uses easy sentences and keeps the list-style of the docs.

debs-obrien:
yes easy sentences are always better. looks good to me.

debs-obrien:
great thanks for your contribution

PR #790, opened 2020-10-02
improve docs

sksharma07774:


manniL:
Hey 👋🏼 
Thanks for the PR but locale is the correct word here. I’m marking the PR as invalid as it lacks effort (no description or reasoning) and doesn’t contain meaningful additions.

Thanks anyway!

PR #803, opened 2020-10-05
Update nuxt-content styling

ilijanovic:
Scoped styles are possible with deep selectors


debs-obrien:
great thanks. I modified the wording a little bit.

PR #837, opened 2020-10-14
fix: improve get started

debs-obrien:


content/en/guides/get-started/installation.md
```diff
@@ -8,29 +8,71 @@ csb_link: https://codesandbox.io/embed/github/nuxt-academy/guides-examples/tree/
...
+```
 
-### Node
+  </code-block>
 
-[node](https://nodejs.org/en/download/) - at least v8.9.0
+</code-group>
 
-_We recommend you have the latest version installed._
+It will ask you some questions (name, Nuxt options, UI framework, TypeScript, linter, testing framework, etc.), when answered, it will install all the dependencies. The next step is to navigate to the project folder and launch it:
```
manniL, line 46: Maybe it'd be good to roughly go through each option and explain what they are there for without going too deep?

debs-obrien: we removed this as it was too difficult to update as it keeps changing and it is clear enough in the create nuxt app which it links to and all info is there

manniL: > we removed this as it was too difficult to update as it keeps changing 

Isn't this rather a bad sign? Rule of thumb: If it is hard to document, it is difficult for the user as well.

> and it is clear enough in the create nuxt app which it links to and all info is there

Maybe it's worth adding a link again, e.g. "Find out more about each question step here"?

debs-obrien: it's more about not having duplicate content to maintain. At the end of the day people just want to run create nuxt app and get an app up and running. I did create a video on it which was 15 mins long and goes over what each one is. This is too much to document as it will scare people away. We could always add the video into the documentation or the link like you suggested - Find out more......

content/en/guides/get-started/installation.md
```diff
@@ -8,29 +8,71 @@ csb_link: https://codesandbox.io/embed/github/nuxt-academy/guides-examples/tree/
 
 ## Prerequisites
 
-Here, you will find information on setting up and running a Nuxt.js project in 4 steps.
+- [node](https://nodejs.org/en/download/) - at least v8.9.0 _We recommend you have the latest version installed._
```
pi0, line 11: ~> It is better to suggest 10 as minimum

pi0: ~> latest **LTS** version

content/en/guides/get-started/installation.md
```diff
@@ -8,29 +8,71 @@ csb_link: https://codesandbox.io/embed/github/nuxt-academy/guides-examples/tree/
...
+- Text editor: Use whatever you like, but we recommend [VSCode](https://code.visualstudio.com/).
```
pi0, line 12: Webstorm also has a really good support for nuxt.js. We can maybe mention both? There is no recommendation point

content/en/guides/get-started/installation.md
Atinux: Updated

content/en/guides/get-started/installation.md
Atinux: Agree, will add a link to WebStorm

content/en/guides/get-started/installation.md
manniL: As long as there is some "Find out more about all the options" link I'm fine with it. Reducing redundant content is good as long as the content is still accessible for the read/user/... 👍

content/en/guides/get-started/installation.md
```diff
@@ -8,37 +8,79 @@ csb_link: https://codesandbox.io/embed/github/nuxt-academy/guides-examples/tree/
 
 ## Prerequisites
 
-Here, you will find information on setting up and running a Nuxt.js project in 4 steps.
+- [node](https://nodejs.org) - at least v10.13 _We recommend you have the latest version installed._
```
pi0, line 11:
```suggestion
- [node](https://nodejs.org) - at least v10.13 _We recommend you have the latest LTS version installed._
```

content/en/guides/get-started/installation.md
debs-obrien: added

pi0 (approved):

PR #839, opened 2020-10-14
refactor(sponsors): improve ref handling

manniL:
As `?ref=nuxt` should be part of all URLs, I'd hardcode it in the `a` tag `href` value instead of adding it to each url manually.


debs-obrien:
cool, thanks

PR #840, opened 2020-10-14
docs(import error note): Note on fixing import error

MentalGear:
Fix for "Cannot use import statement outside a module" error by adding plugins to the transpile build step.
Source: https://stackoverflow.com/a/61196076

### DIFF
--- a/content/en/guide/plugins.md
+++ b/content/en/guide/plugins.md
@@ -41,6 +41,16 @@ Then we can use it directly in our page components:
   }
 </script>
 ```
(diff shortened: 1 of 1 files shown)


debs-obrien:
This is a contribution to the old docs which will disappear soon. @Atinux let me know if you want it added to the new docs

MentalGear:
@Atinux I would think a note on this issue would make sense, since I wasn’t the only one who wasted hours figuring out what the issue was.

Atinux (approved):

debs-obrien:
Thanks @MentalGear. Have added it to the new docs too as you can see from pr above

PR #848, opened 2020-10-15
Switch from `node-sass` to `sass`

MartijnCuppens:
Dart sass (`sass`) is the primary implementation of Sass, is more lightweight and is maintained a little better.

Also, `sass-loader` recommends the usage of `sass`:
https://www.npmjs.com/package/sass-loader


danielroe:
I'd recommend installing `fibers` in this case for better performance.

Atinux:
What would be the example using `fibers`  @danielroe ?

danielroe:
@atinux nothing needs to be done to use `fibers` other than to have it installed in your dependencies. So it would just mean adding this line in each case where `sass` is installed:

```
(It is also recommended to install `fibers` as a devDependency to speed up CSS compilation.)
```

Or, more strongly, you could simply amend the install commands to (for example):
```
 npm install --save-dev sass sass-loader fibers
```

Atinux:
Thanks @danielroe 

I think we can directly suggest:

```bash
npm install --save-dev sass sass-loader fibers
```

And stating that `fibers` is aimed to speed up the compilation process.

Do you mind updating it @MartijnCuppens ?

MartijnCuppens:
> Do you mind updating it @MartijnCuppens ?

Of course not. I've just added `fibers`.

debs-obrien:
Thanks guys 😃

debs-obrien (approved):

febrihidayan (approved):

mrleblanc101:
The fibers documentation did not help me understand why it was needed, but sass docs mention this package.
Maybe some explaination of why fibers is needed should be added to the Nuxt docs too.

PR #858, opened 2020-10-17
contribution.md

ChanderParwani:


manniL:
I'm afraid that this PR doesn't add any substantial value. Thus, I will close it and mark it as `invalid`

PR #880, opened 2020-10-21
Feat/modules link

debs-obrien:


pi0:
@debs-obrien Correct way for PR title is `feat: modules link` :)

debs-obrien:
ooops. how did that happen

manniL:
We might want to introduce semantic pr bot here too (like we have in the core repo)

pi0:
@manniL It is not essential i guess since docs don't have a changelog but would be nice if stick with it 😸

PR #881, opened 2020-10-21
fix: link for module list

manniL:


pi0:
@manniL Do you mind also updating https://github.com/nuxt/nuxtjs.org/blob/master/content/en/api/configuration-modules.md ?

debs-obrien:
oh wow. am sure autosave or something auto added that as I didn't 😡
merging this to fix it and we can create a new pr for old docs if you still think it is necessary but I haven't been adding anything to old docs as they will be removed at end of month @pi0

pi0:
@debs-obrien It will be 404? :O

debs-obrien:
what will be a 404. tested it locally
https://github.com/nuxt/nuxtjs.org/pull/883

pi0:
I meant link to old docs in one month :D (i guess not)

debs-obrien:
ah we will fix that with redirects. it will be a lot of work but worth it as managing 2 sets of docs is a nightmare

PR #879, opened 2020-10-21
docs(en): Correct spelling mistakes

joemirizio:
Fixes spelling mistakes in the English guides.

There were instances of both British English and US English, so I left them as-is. The `nuxt.js` codebase appears to use US English, so I can make the following changes if desired:
- Customising ➡️ Customizing
- artefacts ➡️ artifacts
- behaviour ➡️ behavior

PR #887, opened 2020-10-22
fix: Convert en-GB to en-US

joemirizio:
Per https://github.com/nuxt/nuxtjs.org/pull/879#issuecomment-713440226, this PR converts all en-GB spelling to en-US.

PS: It looks like the `pre-commit` hook reformatted some Markdown tables and added/removed blank lines.


content/en/guides/features/loading.md
```diff
@@ -68,7 +68,7 @@ questions:
 
 Out of the box, Nuxt.js gives you its own loading progress bar component that's shown between routes. You can customize it, disable it or even create your own loading component.
 
-## Customising the Progress Bar
+## Customizing the Progress Bar
```
joemirizio, line 71: This change would break old links pointing to https://nuxtjs.org/guides/features/loading#customising-the-progress-bar. Is that okay?

Atinux: Could you check is we link to it in the docs and update it please?

debs-obrien: I think he means any links that are linking from google results but I don't think its an issue with the anchor tag and if it is we can add a redirect in netlify to cover it but personally I don't think we need to and just need to update any links across the site.

joemirizio: Exactly, only the hash would be different. Probably not a big deal, but I wanted to mention. 

There are no other occurrences of "customising" in this repo, so no links need to get updated.

debs-obrien:
Thanks @joemirizio

PR #890, opened 2020-10-23
docs(en): add runtime config

debs-obrien:
added runtime config to docs as was only in blog post and config glossary. 

Removed env from directory structure docs. let me know if you prefer we keep it in docs and I can put it back. It is still in glossary docs. @Atinux @pi0


Atinux (changes requested):
I think adding a note about the difference of private and public is important. As well as in which context you can access each (server vs server and client)

Atinux (approved):

content/en/guides/directory-structure/nuxt-config.md
```diff
@@ -98,24 +98,6 @@ See more on the [dev property](/guides/configuration-glossary/configuration-dev
 
 </base-alert>
 
-### env
```
pi0, line 101: At least current version of nuxt didn't deprecate `env` option and it has it's purpose of inlining environment variables (We also auto add `NUXT_ENV_*` vars).

I think it worth to be kept in docs with some explanation about use-case and difference (not sure if any other page exists regarding `env` option or not)

content/en/guides/directory-structure/nuxt-config.md
```diff
@@ -225,6 +207,67 @@ See more on the [plugins property](/guides/configuration-glossary/configuration
...
+```html{}[pages/index.vue]
+<script>
+  asyncData ({ $config: { baseURL } }) {
+    const posts = await fetch(`${baseURL}/posts`)
+      .then(res => res.json())
+  }
+</script>
+```
+
+Inside your templates you can access your runtimeConfigs directly using `$config.*`
```
pi0, line 251: Maybe worth mentioning that regardless, if rendering `$config.token` it will be displayed in page (someone had this question in one issue)

content/en/guides/directory-structure/nuxt-config.md
debs-obrien: ok improved this @pi0 let me know if it is better or you want to add further explantion

content/en/guides/directory-structure/nuxt-config.md
debs-obrien: warning added .let me know if it needs further explanation

content/en/guides/directory-structure/nuxt-config.md
pi0: Seems awesome 👍

pi0 (dismissed):

content/en/guides/directory-structure/nuxt-config.md
```diff
@@ -100,16 +100,85 @@ See more on the [dev property](/guides/configuration-glossary/configuration-dev
...
+
+```html{}[pages/index.vue]
+<template>
+  <p>Our Url is: {{ $config.baseURL}}</p>
+</template>
+```
+
+<base-alert type="warn">
+
+You can expose server/private tokens if using `$config` in fetch/asyncData or directly inside your template.
```
pi0, line 166:
```suggestion
It is possible that private config to be exposed to client when using `$config` in `fetch`, `asyncData` or directly inside templates. 
```

/cc @danielroe @manniL  if this can be reworded better

danielroe:
```suggestion
Your private config could be exposed if you use `$config` outside of a server-only context (for example, if you use `$config` in `fetch`, `asyncData` or directly inside your template).
```

PR #912, opened 2020-10-28
生成在打包后自动启动

barry1102:


febrihidayan (approved):

debs-obrien:
thanks for your contribution. These docs have now been archived and are not in use anymore. We would love you to contribute to the new docs instead. That would be amazing

Issue #900, opened 2020-10-26
H2 link bugs


hecktarzuli:
1) H2's show the link icon when you hover, but when you move your mouse to the left to click on the link (to update the url) the link disappears.  You have to wiggle your mouse left/right just the right way to get the anchor image to stay.

2) When you finally click the anchor link and give it to someone, when they load the hash url, it doesn't scroll to the position in the url

It doesn't scroll to these sections when clicking on these links.
https://nuxtjs.org/blog/understanding-how-fetch-works-in-nuxt-2-12#conclusion
https://nuxtjs.org/blog/understanding-how-fetch-works-in-nuxt-2-12#5-error-handling

Notice the funky mouse-over link battle
[nuxt-anchor-bug.mov.zip](https://github.com/nuxt/nuxtjs.org/files/5439771/nuxt-anchor-bug.mov.zip)

PR #922, opened 2020-10-31
fix: icon-link clickable area

xanzhu:
Adds more clickable room around the link icon, making it easier to click! 👀

Reference https://github.com/nuxt/nuxtjs.org/issues/900


debs-obrien:
great, thanks 😃

PR #928, opened 2020-11-02
Fix imgbot automatic SVG optimisation issue

Krutie:
- Added `.imgbotconfig` in the root to ignore all .svg files
- Replaced auto-optimised (and broken) `plugins.svg` image with the original image


debs-obrien:
@Krutie maybe best to only ignore those in the guides/ folder. Other svg's don't seem to have been a problem and this way we are only ignoring your svgs and not across the whole site.

PR #942, opened 2020-11-04
docs(en): Add section about nuxt webpack command

honzabilek4:
Hello, 
I was looking for a command to inspect webpack configuration and found out that it's missing from the documentation. 
It would be really helpful to mention it here, at the point where you're describing how to extend the config. So other people don't have to look it up hard. 
Please take it as a suggestion, all comments are welcomed 🙂 
Thanks!


debs-obrien:
Thanks for your contribution @honzabilek4. I agree this needs to be added to the docs but I prefer to not link to a pr and instead document anything that is necessary. Fell free to work it to add any of the necessary information and if you don't have time I will look into it next week and also see if it needs to be anywhere else in docs. cc @Atinux

honzabilek4:
Tank you @debs-obrien! I thought that might be an issue. I'll try to add more details then 🙂

debs-obrien (changes requested):
add more details instead of pr link

Atinux (approved):

debs-obrien:
merging this and if we get round to adding more info instead of link to pr then we can always improve later. thanks for your contribution

honzabilek4:
Thank you @debs-obrien. I haven't had much time to look into this yet, but I will open a new PR with the improvements soon.

PR #974, opened 2020-11-10
fix: remove typing mistake

miguelalcala-github:


content/en/blog/creating-blog-with-nuxt-content.md
```diff
@@ -288,8 +288,6 @@ All other tags that come from our YAML front matter can be styled as normal eith
 
 To use scoped styles with the nuxt-content class you need to use a deep selector: `/deep/`, `::v-deep` or `>>>`
 
-All other tags that come from our YAML front matter can be styled as normal either using [TailwindCSS](https://tailwindcss.com/) or adding css in the style tag.
```
debs-obrien, line 291: This is actually correct and needs to remain in here. You can style anything that you add in your YAML just like you would normal html. 
content/hello.md
```
---
title: my title
---
```
pages/index.vue
` <h1 class="title">{{ page.title }} </h1>`

debs-obrien (changes requested):
Thanks for your contribution. please amend pr to not remove that paragraph.

content/en/blog/creating-blog-with-nuxt-content.md
miguelalcala-github: Thank you for your reply. Actually I removed it because it is a duplicated paragraph. Isn’t it? Sorry I missunderstand it if it isn’t.

debs-obrien: ahhh ok. I see. It got added again somehow. Ok perfect. Always good to explain why you are removing something as it is not always visible in the pr diff especially when the duplicated content is not directly above/below each other. I suggest in future using the description box to explain what you are removing/editing/amending things.

Thanks for your contributions

PR #1002, opened 2020-11-18
docs(ja): translate configuration-transition

aytdm:
please review this PR.

Ref: https://github.com/nuxt/nuxtjs.org/blob/master/content/ja/_archives/api/configuration-transition.md

resolve vuejs-jp#139


inouetakuya (approved):
Perfect 💯

PR #1025, opened 2020-11-24
feat: add alias section to nuxt-config page

vhoyer:
I don't know if this documentation is correct, please help :fearful:

### DIFF
--- a/content/en/guides/configuration-glossary/configuration-alias.md
+++ b/content/en/guides/configuration-glossary/configuration-alias.md
@@ -1,10 +1,12 @@
 ---
 title: 'The alias Property'
-description: Provide alias for javascript, template, and style
+description: Nuxt.js allows you to use aliases to access custom directories within your JavaScript and CSS
 menu: alias
 category: configuration-glossary
 position: 0
 ---
+
+> Nuxt.js allows you to use aliases to access custom directories within your JavaScript and CSS.
 
 - Type: `Object`
 - Default:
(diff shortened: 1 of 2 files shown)


danielroe (commented):
Great idea to add this. I would suggest:

1. this should be a new page: `~/content/en/guides/configuration-glossary/configuration-alias.md`
2. it should include information on what the default aliases are
3. ...and how these aliases are accessed within JS & CSS

Let me know if you need me to provide content.

vhoyer:
> this should be a new page: ~/content/en/guides/configuration-glossary/configuration-alias.md

but we will keep this description on the nuxt-config page, right? we will add another one of those links like "see more of alias property", right? Cuz I don't think it's a great idea to keep this "hidden" from the main page, sorry if I misinterpreted your words tho :sweat_smile: 

> it should include information on what the default aliases are

Ok, at the time of writing this PR I knew where to look for this, now I don't know anymore ahha, if you know it from the top of your head, I'd appreciate the information, if you would have to search it too, I don't mind doing it them :sweat_smile:. Are the default aliases `srcDir` and `rootDir`?

> ...and how these aliases are accessed within JS & CSS

Ok... how do we access this aliases? I just know about them working on the "import" statement for the CSS loaders :thinking: 

To be 100% sincere, I didn't know if those aliases work on js or template, I just assumed they did, that's why I added my comment on the PR body: "I don't know if this documentation is correct, please help :fearful:"

so please help :fearful: :sweat_smile:

Atinux:
I do believe this is the right move, like @danielroe said it would be nice to have the page for the configuration glossary + an example of using a alias (works in both CSS and JS).

danielroe (approved):

vhoyer:
thanks @danielroe :sob: :heart:

PR #1035, opened 2020-11-28
fix: correct url path -> $config.ngrok.url

lredoban:
Also added the fix to https://github.com/nuxt-academy/examples/pull/1


debs-obrien:
thanks

Issue #1038, opened 2020-11-28
link to Build dir shows 500 error on doc


ryosukemitake:
when you are on the page [here](https://ja.nuxtjs.org/docs/2.x/configuration-glossary/configuration-build), almost all of link in configuration glossary shows 500 errors

## Reproduction
1. go to https://ja.nuxtjs.org/docs/2.x/configuration-glossary/configuration-build
2. click the link buildDir( or other links in configuration glossary menu) from the link on the bottom of the articles or menu on the left
<img width="1255" alt="スクリーンショット 2020-11-29 1 20 20" src="https://user-images.githubusercontent.com/55892311/100520447-0e2a7c00-31e1-11eb-814b-a02b7197f109.png">
3. it hits 500 errors
<img width="1280" alt="スクリーンショット 2020-11-29 1 20 37" src="https://user-images.githubusercontent.com/55892311/100520452-18e51100-31e1-11eb-8ff9-fc95ac77d5f2.png">

What is Expected?
link takes you to the expected articles

debs-obrien:
Hi. Thanks for reporting. Loving the Nuxt in pink 😃. Yes this is because translation is still a work in progress so those pages don't exist and that gives a 500 error. 

@inouetakuya @aytdm what do you think about adding those missing pages in English to avoid this error and then translate them when you have time as right now there is no fallback to English so if it doesn't exist it will cause an error. Thanks

inouetakuya:
> what do you think about adding those missing pages in English to avoid this error and then translate them when you have time as right now there is no fallback to English so if it doesn't exist it will cause an error.

It sounds good 👍 

I will create the pull request 💪

inouetakuya:
For example the following page has no translated page and fallback to English content now.  Fallback is good but it has no link on left side menu so it is still a good idea to add missing pages in English to content/ja/ ☺️ 

https://ja.nuxtjs.org/docs/2.x/configuration-glossary/configuration-router/

aytdm:
I've merged the PR (#1041). Thank you for your report and suggestions :)

PR #1041, opened 2020-11-29
docs(ja): add missing pages in English to content/ja/guides/

inouetakuya:
@aytdm 

ref: https://github.com/nuxt/nuxtjs.org/issues/1038#issuecomment-735286190

I have copied the missing pages in content/ja/guides from content/en/guides to fulfill the links on the left side menu so that it would be easier to access to other pages.


aytdm (approved):
Thanks

PR #1057, opened 2020-12-03
docs(en) documentation on eslint-module v2 to v3 upgrade.

rlam3:
Adding documentation on eslint-module v2 to v3 upgrade.


content/en/faq/extend-webpack.md
```diff
@@ -60,6 +60,36 @@ export default {
...
+export default {
+  build: {
+    extend(config, { isDev, isClient }) {
+      // Run ESLint on save
+      if (ctx.isDev && ctx.isClient) {
+        const options = {
+          exclude: ['node_modules']
+        }
+        const EslintPlugin = require('eslint-webpack-plugin')
+        config.plugins.push(new EslintPlugin(options))
```
ricardogobbosouza, line 92: The eslint module already makes this configuration, it makes more sense to recommend it

https://github.com/nuxt-community/eslint-module/blob/master/lib/module.js#L40

content/en/faq/extend-webpack.md
```diff
@@ -60,6 +60,36 @@ export default {
           loader: 'eslint-loader',
           exclude: /(node_modules)/
```
ricardogobbosouza, line 61: I think we should remove this configuration using the `eslint-loader`

ricardogobbosouza (changes requested):

Atinux:
I think this should be a documentation about using https://github.com/nuxt-community/eslint-module

debs-obrien:
Just to be clear. Should we remove this section of the FAQ page:
Screenshot 2020-12-03 at 17 58 31

https://nuxtjs.org/faq/extend-webpack

debs-obrien:
> I think this should be a documentation about using https://github.com/nuxt-community/eslint-module

where do you suggest we place this @Atinux. We did have a section on dev tooling before but it got removed. We could add another one with suggestions of dev tools to use to improve your Nuxt experience and list the modules with a short description. What do you think?

rlam3:
> > I think this should be a documentation about using https://github.com/nuxt-community/eslint-module
> 
> where do you suggest we place this @Atinux. We did have a section on dev tooling before but it got removed. We could add another one with suggestions of dev tools to use to improve your Nuxt experience and list the modules with a short description. What do you think?

I think we should include this still include this in the docs because a new project I create with nuxt still requires v2 of eslint-module. It isn't created with v3. Unless we can have an option to choose v2 vs v3 upon creation. It is also a great reference point as to when and where users of these modules need to take notice of updating their configuration files.

Personally, I believe it is better for documentation to guide users of this module to update correctly, instead of being forced into updates lacking documentation. I too was confused as to why my app kept breaking due to a major update. 

The following is what a new Nuxt Project for me looked like. 

```js
//package.json
{
  "name": "newN2020",
  "version": "1.0.0",
  "private": true,
  "scripts": {
    "dev": "nuxt-ts",
    "build": "nuxt-ts build",
    "start": "nuxt-ts start",
    "generate": "nuxt-ts generate",
    "lint:js": "eslint --ext .js,.vue --ignore-path .gitignore .",
    "lint": "npm run lint:js",
    "test": "jest"
  },
  "lint-staged": {
    "*.{js,vue}": "eslint"
  },
  "husky": {
    "hooks": {
      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
      "pre-commit": "lint-staged"
    }
  },
  "dependencies": {
    "@nuxt/content": "^1.9.0",
    "@nuxt/typescript-runtime": "^2.0.0",
    "@nuxtjs/axios": "^5.12.2",
    "@nuxtjs/pwa": "^3.0.2",
    "core-js": "^3.6.5",
    "nuxt": "^2.14.6"
  },
  "devDependencies": {
    "@commitlint/cli": "^11.0.0",
    "@commitlint/config-conventional": "^11.0.0",
    "@nuxt/types": "^2.14.6",
    "@nuxt/typescript-build": "^2.0.3",
    "@nuxtjs/eslint-config": "^3.1.0",
    "@nuxtjs/eslint-config-typescript": "^3.0.0",
    "@nuxtjs/eslint-module": "^2.0.0", <<<<<<<<<<<<<<<<<<<<
    "@nuxtjs/tailwindcss": "^3.1.0",
    "@vue/test-utils": "^1.1.0",
    "babel-core": "7.0.0-bridge.0",
    "babel-eslint": "^10.1.0",
    "babel-jest": "^26.5.0",
    "eslint": "^7.10.0",
    "eslint-config-prettier": "^6.12.0",
    "eslint-plugin-nuxt": "^1.0.0",
    "eslint-plugin-prettier": "^3.1.4",
    "husky": "^4.3.0",
    "jest": "^26.5.0",
    "lint-staged": "^10.4.0",
    "prettier": "^2.1.2",
    "ts-jest": "^26.4.1",
    "vue-jest": "^3.0.4"
  }
}


```

Uses v2 instead of v3

```
❯ npx npm-check -u
 Major Update Potentially breaking API changes. Use caution.
 ◯ @nuxtjs/eslint-config devDep             3.1.0  ❯  5.0.0  https://github.com/nuxt/eslint-config/tree/master/packages/eslint-config
 ◯ @nuxtjs/eslint-config-typescript devDep  3.0.0  ❯  5.0.0  https://github.com/nuxt/eslint-config/tree/master/packages/eslint-config-typescript
 ◯ @nuxtjs/eslint-module devDep             2.0.0  ❯  3.0.1  https://github.com/nuxt-community/eslint-module#readme
 ◯ eslint-plugin-nuxt devDep                1.0.0  ❯  2.0.0  https://github.com/nuxt/eslint-plugin-nuxt#readme
```

content/en/faq/extend-webpack.md
rlam3: @ricardogobbosouza Thank you for the great work for updating to use this package. I want some extra clarification on this package if possible. 

What is the order of the options being applied? Which is overriding what? There are so many levels of options that I am confused as to which I should really be targeting now. I was at first assuming we should leave all options inside the .eslintrc.js file but it seems like you also recommend top-level eslint options within the nuxt.config.js.

So my question is which eslint options take precedence? 
I apologize if the write-up may sound confusing.

ricardogobbosouza:
Hi @ rlam3
Are you using https://github.com/nuxt/create-nuxt-app?

rlam3:
> Hi @rlam3
> Are you using https://github.com/nuxt/create-nuxt-app?

@ricardogobbosouza yes, i am using the `npx create-nuxt-app <projectName>` cli command. Am I missing something?

ricardogobbosouza:
@rlam3 
I'll try to explain...

**V2**
The [eslint-module](https://github.com/nuxt-community/eslint-module/tree/v2.0.0) in version `2` does exactly this configuration using the deprecated `eslint-loader`, see https://github.com/nuxt-community/eslint-module/blob/v2.0.0/lib/module.js#L32

**V3**
The [eslint-module](https://github.com/nuxt-community/eslint-module/tree/v3.0.0) in version `3`, use the` eslint-webpack-plugin`, see https://github.com/nuxt-community/eslint-module/blob/master/lib/module.js#L36

In theory just update the module to v3

I will release a new version of `eslint-module` 3.0.2 by updating the `eslint-webpack-plugin` which is in version `^2.4.1` which solves some bugs

So here in the **NuxtJS** documentation it makes more sense to recommend the [eslint-module](https://github.com/nuxt-community/eslint-module) so that the user does NOT need to make this configuration manually, in addition to the module containing other benefits.

**We can simply recommend the [eslint-module](https://github.com/nuxt-community/eslint-module#setup), therefore, this configuration below is not necessary**

```js
// eslint-loader

export default {
  build: {
    extend(config, { isDev, isClient }) {
      if (isDev && isClient) {
        config.module.rules.push({
          enforce: 'pre',
          test: /\.(js|vue)$/,
          loader: 'eslint-loader',
          exclude: /(node_modules)/
        })
      }
    }
  }
}

// eslint-webpack-plugin
export default {
  build: {
    extend (config, {isDev, isClient}) {
      // Run ESLint on save
      if (ctx.isDev && ctx.isClient) {
        const options = {
          exclude: ['node_modules']
        }
        const EslintPlugin = require ('eslint-webpack-plugin')
        config.plugins.push (new EslintPlugin (options))
    }
  }
}
```

**We can simply recommend the eslint-module**

I hope I have clarified 😄

PR #1064, opened 2020-12-05
fix(content): nuxt class link

zeravcic:
The `Nuxt Class` link in the 'Core > Nuxt' section of the 'Nuxt Modules Intro' page was incorrectly pointing to the [Using Nuxt.js Programmatically](https://nuxtjs.org/docs/2.x/internals-glossary/nuxt) page.

This change updates the link target from `/docs/2.x/internals-glossary/nuxt` to `/docs/2.x/internals-glossary/internals-nuxt`.


debs-obrien:
Great. Thanks 😊

PR #1067, opened 2020-12-07
docs(en) Proposed changes `addPlugin` API

gangsthub:
As of current implementation of https://github.com/nuxt/nuxt.js/blob/dev/packages/core/src/module.js

The object syntax for `addPlugin(template)` was missing.

### DIFF
--- a/content/en/guides/internals-glossary/internals-module-container.md
+++ b/content/en/guides/internals-glossary/internals-module-container.md
@@ -57,7 +57,7 @@ This method returns final `{ dst, src, options }` object.
 
 - **template**: On top of [`addTemplate`](#addtemplate-template)'s `template` Object properties (`src`, `options`, and `fileName`),
 the parameter `template` in this scope also has plugin-specific options:
-  - `mode` (`'client'` or `'server'`): as in plugins [mode](/docs/2.x/directory-structure/plugins#object-syntax) 
+  - `mode` (`'all'`, `'client'` or `'server'`): as in plugins [mode](/docs/2.x/directory-structure/plugins#object-syntax) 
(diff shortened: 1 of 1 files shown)


content/en/guides/internals-glossary/internals-module-container.md
```diff
@@ -55,6 +55,11 @@ This method returns final `{ dst, src, options }` object.
 
 ### addPlugin (template)
 
+- **template**: On top of [`addTemplate`](#addtemplate-template)'s `template` Object properties (`src`, `options`, and `fileName`),
+the parameter `template` in this scope also has plugin-specific options:
+  - `mode` (`'all'`, `'client'` or `'server'`): as in plugins [mode](/docs/2.x/directory-structure/plugins#object-syntax) 
+  - `ssr` (`Boolean`): Prefer `mode`, as this option will be deprecated in Nuxt 3.
```
pi0, line 61: Neither ssr or mode options are not recommended. Mode is better be part of plugin name like `axios.client.js`

gangsthub: Ah, good to know. I'll change the content, then ;) Thanks, Pooya!

gangsthub:
@pi0 I edited the `addPlugin` section again, trying to follow your review. Please, feel free to edit it directly or, if you want, suggest any changes. Thank you!

pi0:
Thanks @gangsthub. I've modified it to be similar to [plugins](https://nuxtjs.org/docs/2.x/directory-structure/plugins#client-or-server-side-only) page but mentioning name convention. We can officially deprecate when API for nuxt3 is finalized with module kit (replacing built-in module container and self documented)

pi0 (approved):

PR #1080, opened 2020-12-13
docs(ja):  fix name of example to match with object listed above

ekrecker:
I've resolved [vuejs-jp#53](https://github.com/vuejs-jp/ja.nuxtjs.org/issues/53)


aytdm (approved):
Thanks

PR #1085, opened 2020-12-15
ci: check links on push to `master` and allow manual trigger

danielroe:


danielroe:
It's currently complaining because not all languages have the target pages of some redirects - I can fix this, but that depends on whether they *should* have these links or not. Perhaps allow redirects to fail on languages other than nuxtjs.org?

debs-obrien:
Thats a good question. I think not fail on other languages but show error is helpful so if there was a warning for them then it might help translators fix them. But we shouldn't fail our site for errors of other langs especially when they are not always uptodate or complete.

PR #1105, opened 2020-12-24
style: some suggested tweaks

danielroe:
* use their preferred names/capitalisation for companies/projects
* where multiple options are valid, standardise on the most common within docs
* use US rather than UK English
* prefer `e.g.` to `eg.`
* fix some typos


Atinux (approved):

Atinux:
Thanks @danielroe 💚

PR #1116, opened 2020-12-31
docs(zh) update translation：content/zh/guides/get-start

thedb:
Updated this part: get-started

📝 guides -> get-started
 - [x]  commands
 - [x]  conclusion.md
 - [x]  directory-structure.md
 - [x]  installation.md
 - [x]  routing.md
 - [x]  upgrading.md

### DIFF
--- a/content/zh/guides/get-started/commands.md
+++ b/content/zh/guides/get-started/commands.md
@@ -1,0 +1,193 @@
+---
+title: 命令与部署
+description: Nuxt.js自带一组有用的命令，用于部署开发和生产环境。
+position: 4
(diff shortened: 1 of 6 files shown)


debs-obrien:
great thanks. although this will be merged you won't see it live until we add 'zh' to array in guides-slug. Once we do that we will loose the 'en' fallback so best not to add this until we have a good majority translated. We can then make them live and if there are some chapters missing we can copy and paste the 'en' of these into the 'zh' folder until they are translated. But for now lets keep working the way you are and am here if you need any help. Thanks 💪

PR #1122, opened 2021-01-04
Fix(ja): parse errors

debs-obrien:


aytdm (approved):
Thank you. I'll keep that can't use backtick in description in mind.

inouetakuya (approved):
Thank you!

> I'll keep that can't use backtick in description in mind.

me, too.

PR #1146, opened 2021-01-14
Fix typo in french documentation

bokub:
Replace "montages" with "montag**n**es" (which means "mountains" :mountain: ).


debs-obrien:
merci

PR #1149, opened 2021-01-14
Update `terserOptions` documentation

honzabilek4:
This part took me a while to figure out when dealing with source maps in Nuxt. This was mainly due to the misleading content in the updated paragraph. 
Here's my attempt to make it more clear and save other devs some precious time.


content/en/guides/configuration-glossary/configuration-build.md
```diff
@@ -727,7 +727,7 @@ Templates are rendered using [`lodash.template`](https://lodash.com/docs/#templa
 
 Terser plugin options. Set to `false` to disable this plugin.
 
-`sourceMap` will be enabled when webpack `config.devtool` matches `source-?map`
+Enabling `sourceMap` will leave `//# sourceMappingURL` linking comment at the end of each output file if webpack `config.devtool` is set to `source-map`.
```
pi0, line 730: For reference this is [terser-webpack-plugin](https://github.com/webpack-contrib/terser-webpack-plugin/blob/v4.2.3/src/index.js#L605) logic which is removed for v5 but we won't have it for nuxt2 (since it dropped webpack4...)  @clarkdo

pi0 (approved):
Thanks!

PR #1172, opened 2021-01-19
ci: fix cypress tests (add trailing slash)

### LINKED ISSUE #1171: end2end tests failing
end2end tests are failing since this pr: https://github.com/nuxt/nuxtjs.org/pull/1160/files which includes updating Nuxt to v2.15 using nuxt-edge and adding esm to bundle so crawler works. 

Tests are running fine locally but failing in CI. See failing tests here:https://github.com/nuxt/nuxtjs.org/runs/1722868741

Error message:
CypressError: `cy.visit()` failed trying to load:

http://localhost:3000/docs/2.x/get-started/installation

The response we received from your web server was:

  > 404: Not Found

This was considered a failure because the status code was not `2xx`.

danielroe:
(Trailing slash behaviour for local `nuxt start` changed in https://github.com/nuxt/nuxt.js/pull/8398.)

closes #1171

### DIFF
--- a/cypress/integration/home.spec.js
+++ b/cypress/integration/home.spec.js
@@ -6,7 +6,7 @@ describe('The Home Page links', () => {
     cy.get('a[data-cy="get-started"]')
       .click()
     cy.url().should('include', '/docs/2.x/get-started/installation')
-    cy.visit('/docs/2.x/get-started/installation')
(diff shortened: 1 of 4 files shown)


debs-obrien:
Thanks @danielroe

PR #1174, opened 2021-01-19
feat: reduce faq

debs-obrien:
reduce faq by removing duplicate content or adding content to configuration as shouldn't be in faq
redirect to module for module faq as easier for maintainence
update redirects
old faq moved to archives incase we still need it


debs-obrien:
Awaiting feedback from @pi0 on if to include the pages on Auth Routes and Auth External API or if to redirect to auth module for these pages. All other faq pages have been greatly reduced and merged with docs. Now to decide if use faq or tips as name for this section. cc @Atinux 
preview: https://deploy-preview-1174--nuxt.netlify.app/faq

debs-obrien:
faq is now looking very reduced.
cc @Atinux @pi0 
https://deploy-preview-1174--nuxt.netlify.app/faq

PR #1192, opened 2021-01-25
Fixed the typo causing italic instead of bold

rexarvind:
Screenshot_2021-01-25-15-18-47-889

Double asterisks outside the line was causing italic and I moved them inside the square brackets.

Line Fixed:
The application is now running on *[http://localhost:3000](http://localhost:3000/).*


rexarvind:
Screenshot_2021-01-25-17-18-41-665

I think there's problem somewhere else in rendering markdown, that line is still looking the same.

PR #1193, opened 2021-01-25
docs: update runtimeConfig example to show dynamic baseURL

katerlouis:
I may be missing something, but a boolean expression with 2 strings doesn't make a lot of sense, does it? My best guess (and @danielroe in the nuxt discord) was that this example wants to provide different urls depending of production-env or not. Then again why mentioning `.env.example` in the first place? 

So basically this PR is not inteded to provide the best solution, but more to start a conversation.

### DIFF
--- a/content/en/blog/moving-from-nuxtjs-dotenv-to-runtime-config.md
+++ b/content/en/blog/moving-from-nuxtjs-dotenv-to-runtime-config.md
@@ -107,7 +107,7 @@ Also this can be a better replacement for `.env.example` and the default values 
 ```js{}[nuxt.config.js]
 export default {
   publicRuntimeConfig: {
-    baseURL: 'https://dev.nuxtjs.org' || 'https://nuxtjs.org'
+    baseURL: process.env.NODE_ENV === production ? 'https://nuxtjs.org' : 'https://dev.nuxtjs.org'
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


content/en/blog/moving-from-nuxtjs-dotenv-to-runtime-config.md
```diff
@@ -107,7 +107,7 @@ Also this can be a better replacement for `.env.example` and the default values
 ```js{}[nuxt.config.js]
 export default {
   publicRuntimeConfig: {
-    baseURL: 'https://dev.nuxtjs.org' || 'https://nuxtjs.org'
+    baseURL: process.env.NODE_ENV === production ? 'https://nuxtjs.org' : 'https://dev.nuxtjs.org'
```
danielroe, line 110:
```suggestion
    baseURL: process.env.NODE_ENV === 'production' ? 'https://nuxtjs.org' : 'https://dev.nuxtjs.org'
```

clarkdo (approved):

danielroe (approved):

pi0 (approved):

Atinux (approved):

manniL (approved):

debs-obrien:
Thanks @katerlouis

katerlouis:
Woh :D – well you guys act fast ^^
Thanks for accepting my PR. But I'm still wondering why the docs mention `.env.example` when the following code snippet does not use it / refer to it in any way? Since I don't know why it's mentioned, I cannot suggest a better solution. Any ideas?

katerlouis:
I still don't get it :D
but when you say this is correct now the way it is, I'm satisfied. I guess we already corrected the obvious mistake. All I can say as a noobie is that `.env.example` confuses me. :D

PR #1195, opened 2021-01-25
Deploy to digitalocean App Platform guide

Dawntraoz:
Add DO App Platform guide to Deployments


debs-obrien (changes requested):
position should be 104.1 then it will be in alphabetical oder. by adding decimal we don't have to modify all other files

debs-obrien:
Overall Great work @Dawntraoz and nice contribution. Just minor change on ordering above

debs-obrien:
Screenshot 2021-01-25 at 18 07 14
Not sure if should be Digital Ocean as name in menu or DO App Platform. I am not familiar with it shortened to DO but could be just me. @pi0 what do you think?

Dawntraoz:
> Screenshot 2021-01-25 at 18 07 14
> 
> Not sure if should be Digital Ocean as name in menu or DO App Platform. I am not familiar with it shortened to DO but could be just me. @pi0 what do you think?

I added it because in their menu https://www.digitalocean.com/docs/app-platform/how-to/add-deploy-do-button/ they have DO for long titles

debs-obrien:
I see. however there it makes sense to shorten it as you are already on the Digital Ocean platform whereas on our site perhaps it is better to just have Digital Ocean or do you think it is more more important to have App platform in the menu?

Dawntraoz:
> I see. however there it makes sense to shorten it as you are already on the Digital Ocean platform whereas on our site perhaps it is better to just have Digital Ocean or do you think it is more more important to have App platform in the menu?

I think App Platform should be there, because you can still creating a Droplet in DigitalOcean to deploy your site 🤔

debs-obrien (approved):

clarkdo (approved):

PR #1220, opened 2021-02-04
Added just a small info tab about global CSS files

adelriosantiago:
Nuxt makes most of the things automatically so I was thinking that people may think that global CSS files are loaded automatically. Which is not true. This makes it clear.


debs-obrien:
nice. thanks

PR #1226, opened 2021-02-06
docs: update `sass-loader` recommendation to v10

danielroe:
https://github.com/nuxt/nuxt.js/issues/8771


aytdm (approved):
Thank you :)

pi0:
@danielroe Thanks. Also since this might be a common mistake, maybe we should add a check to framework for resolved version of sass-loader

inouetakuya (approved):

PR #1233, opened 2021-02-07
docs: update default webpack chunk filenames

danielroe:
see https://github.com/nuxt/nuxt.js/pull/8778


clarkdo (approved):

pi0:
Note: Part of this change is to be released by 2.15 :)

PR #1237, opened 2021-02-08
docs: use `token` consistently throughout post

suruaku:
One variable name was missing in some places and it was incorrect.


suruaku:
Forgot to mention, authtoken name is already used in that example [last code part](https://nuxtjs.org/blog/creating-a-nuxt-module#full-code-example)

Also env variable won't be used since it is assigned to `const token`, but never used? (lines 232 & 238)

content/en/blog/creating-a-nuxt-module.md
```diff
@@ -219,7 +219,7 @@ export default function () {
 }
 ```
 
-We can then await the ngrok authtoken passing in the value of `token`which we have defined above.
+We can then await the ngrok authtoken passing in the value of `authtoken`which we have defined above.
```
clarkdo, line 222: Let's add a space after `

danielroe (commented):
@suruaku Very good catch on quite a lot of this. My only suggestion is to use `token` for the variable name (as before) to distinguish it from the ngrok method - but to keep the rest of your changes.

suruaku:
Yes sure, using `token` makes more sense after reading this article for many times.

danielroe (approved):

PR #1273, opened 2021-02-17
Fixed hydration error

smaknsk:
Fix hydration error. If there is $ is_args we force to forward to nodejs

If the query string changes the content using a request to the server, then we will receive an error:
"Failed to execute 'appendChild' on 'Node': This node type does not support this method"
https://github.com/nuxt/nuxt.js/issues/5800

### DIFF
--- a/content/en/guides/deployment/nginx-proxy.md
+++ b/content/en/guides/deployment/nginx-proxy.md
@@ -89,6 +89,9 @@ server {
         add_header Strict-Transport-Security "max-age=31536000; includeSubDomains; preload" always;
         add_header X-Frame-Options "SAMEORIGIN";
 
+        error_page 420 = @nodejs;
+        if ( $is_args ) { return 420; }
+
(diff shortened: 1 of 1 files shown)


content/en/guides/deployment/nginx-proxy.md
```diff
@@ -89,6 +89,9 @@ server {
         add_header Strict-Transport-Security "max-age=31536000; includeSubDomains; preload" always;
         add_header X-Frame-Options "SAMEORIGIN";
 
+        error_page 420 = @proxy;
```
clarkdo, line 92: Why 420 is being used here ?

smaknsk: Any non-existent HTTP code for use inside "if". The real code will be passed from "location @proxy"
if ( $is_args ) { return 420; } => 420 code handles @proxy

Issue #1244, opened 2021-02-10
Component discovery docs


pi0:
Currently we have sparse documentation about [nuxt/components](https://github.com/nuxt/components). Since this is an ad-hoc module (built-in feature with independent versioning) i think it would be the best to move and link docs from repo to official website so is easier for users to find all information in single place and avoid content to be stalled.

Goals:

- Adding "Features > Components Discovery" section
 - Renaming "Features > Nuxt Components" to "Features > Components" (or built-in components?)
- Update "Directory Structure > Components" with components v2 docs (to be released)
- Update "Configuration Glossary > components" with detailed version of options (moving from repo)
- [q] Where to put advanced docs about extending components with hooks (module authors) and diagram? We may add it under features?
- Apply and mention conventions based on [vue-style-guide](https://vuejs.org/v2/style-guide/) (for base components, naming, etc)

Related PR: #1363 (not fully covering this issue atm)

pi0:
> I think it could be directly into "Configuration Glossary > components

Hmm but i'm afraid other hooks aren't in configuration docs. This section would be also extended probably for including tips for unit test setup and internals

> Features > Built-in Components

Seems good 👍

pi0:
I would go with a try with internals then :)

PR #1279, opened 2021-02-19
Update documentation about nested directories components

riochandra1212:
Migration nuxt/component v1 to v2 makes a great feature about nested directories, starting `nuxt@2.15` using `@nuxt/component v2`. [nested directories nuxt component](https://github.com/nuxt/components#nested-components)

at first, I got problem at nuxt production that update nuxt/component to v2. Many component that not called directly.

I think nuxt-newbie will get this problem when learn it for the first time, so i decide to edit documentation about this update.

I edit documentation about nuxt [nested directories component](https://nuxtjs.org/docs/2.x/directory-structure/components#nested-directories) relate to new update @nuxt/component version 2.

### DIFF
--- a/content/en/guides/directory-structure/components.md
+++ b/content/en/guides/directory-structure/components.md
@@ -187,6 +187,6 @@ components: {
 And now in your template you can use `AwesomeButton` instead of `BaseFooButton`.
 
 ```html{}[pages/index.vue]
-<FooButton />
+<AwesomeButton />
 ```
 <base-alert type="next">Learn more about the [components module](/blog/improve-your-developer-experience-with-nuxt-components).</base-alert>
--- a/content/en/guides/directory-structure/components.md
+++ b/content/en/guides/directory-structure/components.md
@@ -178,15 +178,15 @@ components: {
     '~/components', // shortcut to { path: '~/components' }
       {
         path: '~/components/base/foo/',
-        prefix: 'awesome'
+        prefix: 'foo'
       }
   ]
 }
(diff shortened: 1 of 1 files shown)


content/en/guides/directory-structure/components.md
```diff
@@ -153,46 +153,40 @@ If you have components in nested directories such as:
...
 ```bash{}[nuxt.config.js]
 components: {
   dirs: [
-    '~/components',
+    '~/components', // shortcut to { path: '~/components' }
       {
-        path: '~/components/base/',
-        prefix: 'Base'
+        path: '~/components/base/foo/',
+        prefix: 'awesome'
```
pi0, line 181: Default prefix is (with directory name) is to ensure fs path matches component name. This way we can quickly find component based on it's name. `prefix` is an advanced option and mostly usable when creating component libraries. For this docs section, i would suggest simplify example by registering `~/components/base` to remove `Base` prefix.

riochandra1212: Ok, that's make sense, i'll fix that

riochandra1212:
@pi0 I've fix default prefix that you mention earlier, may you review it.

pi0:
Thanks @riochandra1212. I've added some changes to explain simplest way to remove prefix. Advanced `prefix` and `pathPrefix` options are explained in longer docs linked in the bottom. We will move all docs into website soon (see #1244)

pi0 (approved):

PR #1281, opened 2021-02-20
fix: remove recursive component reference in `<AdsBlock>`

JeffreyDotDev:
Fix the recursive load of the AdsBlock component

Probably caused by a typo because the component that needs to be loaded seems to be AdBlock, without the extra s.


danielroe (approved):

danielroe:
Thanks @JeffreyDotDev - great catch ❤️

PR #1286, opened 2021-02-23
Update to reflect current behavior

travis5491811:
I updated this to reflect the actual behavior and to reflect the documentation from https://github.com/nuxt/components


danielroe:
Thanks very much for this - good spot! - however it's already under development in https://github.com/nuxt/nuxtjs.org/pull/1279

PR #1298, opened 2021-03-02
docs: add example of refetching on activated if fetch error in blog article

ziaadini:


Atinux:
What do you think @danielroe ?

danielroe:
I think this is in general a good pattern - to check for error and refetch - but I'm not sure that it fits in this part of the blog, which is focusing on performance optimisation.

danielroe (commented):
On reflection, if we were to add something I think we would add it in this section: https://github.com/nuxt/nuxtjs.org/blob/6142abbf202e7479f98a8e1d39b6e598b78040b3/content/en/blog/understanding-how-fetch-works-in-nuxt-2-12.md#L123

But it would need a comment and justification so the reader understands why to refetch in this instance. Why are we refetching?

PR #1305, opened 2021-03-06
docs(ja): fix to add single quotes in directory-structure/plugins

aytdm:
Fix parse error

>  WARN  Could not parse ./content/ja/guides/directory-structure/plugins.md: end of the stream or a document separator is expected at line 3, column 14:
>     description: `plugins` ディレクトリには、ルート Vue.js アプ ...


inouetakuya (approved):
Thank you for everything! 🙏

PR #1306, opened 2021-03-06
docs(ja): translate internals-glossary/nuxt-render-and-get-window

aytdm:
Please review this PR.

ref: https://github.com/nuxt/nuxtjs.org/blob/main/content/ja/_archives/api/nuxt-render-and-get-window.md

resolve vuejs-jp#120
resolve https://github.com/vuejs-jp/ja.nuxtjs.org/issues/286


inouetakuya (approved):
LGTM! 👍

PR #1310, opened 2021-03-06
docs(ja): translate internals-glossary/internals-renderer

aytdm:
Please review this PR.

ref: https://github.com/nuxt/nuxtjs.org/blob/main/content/ja/_archives/api/internals-renderer.md

resolve vuejs-jp#155


inouetakuya (approved):
Thank you for everything! 👍

PR #1320, opened 2021-03-09
docs: add example complex store structure

RobSteward:
Added extended file structure representation for full store use on complex projects. If someone could confirm whether the "modules" folder is still required for modules to work under nuxt (essential store/shop/cart/actions.js vs store/modules/shop/cart/actions.js), that would be great.

### DIFF
--- a/content/en/guides/directory-structure/store.md
+++ b/content/en/guides/directory-structure/store.md
@@ -210,6 +210,38 @@ export default {
 }
 ```
 
+A complex store setup file/folder structure might look as follows (in `store/`)
+
+```
+ store
+   |_modules
+     |_shop
+       |_cart
+         |_actions.js
+         |_getters.js
+         |_mutations.js
+         |_state.js
+       |_products
+         |_itemsGroup1
+           |_actions.js
+           |_getters.js
+           |_mutations.js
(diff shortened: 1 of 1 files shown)


content/en/guides/directory-structure/store.md
```diff
@@ -210,6 +210,38 @@ export default {
...
+         |_actions.js
+         |_getters.js
+         |_mutations.js
+         |_state.js
+   |_index.js (containing plugins and other additional elements, see below)
+   |_actions.js (being the root actions, contains nuxtServerInit)
+   |_getters.js (being the root getters)
+   |_mutations.js (being the root mutations)
+   |_state.js (being the root state)
+```
```
danielroe, line 243: @RobSteward No, you don't need a `modules` directory, and I think this could be simplified a bit and still get the point across. Something like:
```
store/
--| index.js
--| ui.js
--| shop/
----| cart/
------| actions.js
------| getters.js
------| mutations.js
------| state.js
----| products/
------| mutations.js
------| state.js
------| itemsGroup1/
--------| state.js
```

RobSteward: Agreed. Good enough.

PR #1327, opened 2021-03-13
docs(ja): document the unit for the `interval` property

aytdm:
resolve https://github.com/vuejs-jp/ja.nuxtjs.org/issues/163


inouetakuya (approved):
Thank you! 👍

PR #1339, opened 2021-03-14
docs: update node version in github-pages workflow

vedmaque:
Caused by https://github.com/nuxt/nuxt.js/blob/dev/packages/cli/src/utils/dependencies.js#L11

```
const nodeVersion = '>=12.0.0';

if (!semver.satisfies(process.version, nodeVersion)) {
  consola__default['default'].warn(`You are using an unsupported version of Node.js (${process.version}). It is recommended to use the latest LTS version (https://nodejs.org/en/about/releases)`);
}
```


danielroe (approved):

danielroe:
Good catches - thanks! ❤️

PR #1351, opened 2021-03-18
docs: update port/host for google cloud run

nathanfletcher:
Updated to use Google Cloud run default settings for Docker.
I removed the server config settings because they get suppressed by the `NUXT_HOST` and `NUXT_PORT` env variables values for Cloud Run
This seems straightforward and much easier to understand but I'll need reviewers to confirm that.


danielroe (commented):
You're right that this should be amended. However, instead of `NUXT_HOST` and `NUXT_PORT` just `HOST` and `PORT` are sufficient. And defaulting to 8080 is a good call, whilst allowing it to be overridden.

[Docs reference](https://cloud.google.com/run/docs/reference/container-contract#port).

nathanfletcher:
hi @danielroe you know what, you're absolutely right.
I tried it on my Cloud Run instance and it worked. 
I've updated it to reflect the change.

danielroe (approved):

PR #1355, opened 2021-03-20
Add NUXT_PORT: 8080 into env_variables

wakamin:


content/en/guides/deployment/appengine-deployment.md
```diff
@@ -40,6 +40,7 @@ handlers:
 
 env_variables:
   HOST: '0.0.0.0'
+  NUXT_PORT: 8080
```
danielroe, line 43: I don't think this is correct. 

App Engine injects the `PORT` environment variable itself ([see docs](https://cloud.google.com/appengine/docs/standard/nodejs/building-app/writing-web-service#key_points)) - which Nuxt respects automatically. By default it is 8080, but it isn't guaranteed to be.

wakamin: In my case Nuxt run in port 3000 while app engine is run in port 8080. The app engin url cannot be accessible.

Add NUXT_PORT: 8080 to make sure nuxt is run in port 8080.

danielroe: @wakamin Are you sure? In the link I pasted, they say:

> Notice that if process.env.PORT is not set, port 8080 is used as a default. This is necessary for testing your app locally, because `process.env.PORT` doesn't get set during local runs - it is only set when your app is running on App Engine. You can use whichever port you prefer for testing, but this guide uses 8080.

`process.env.PORT` should work just as well as `process.env.NUXT_PORT` and is preferred.

PR #1358, opened 2021-03-22
fix: reenable @nuxt/eslint-module

rootblack45:
Because https://github.com/nuxt-community/eslint-module/pull/46 was fixed and had been released in version 3.0.1. Now in nuxtjs.org is using @nuxtjs/eslint-module version 3.0.2 that we can reenable from now.


danielroe (approved):
Thanks! ❤️

PR #1369, opened 2021-03-27
docs(context-helpers): specify the three boolean values in process

aesyondu:
While reading the [context helpers guide](https://nuxtjs.org/docs/2.x/concepts/context-helpers), it is mentioned that there are three boolean values injected to the global process. However, only one boolean value was mentioned: `process.client`.

I was curious what the other two were, so I viewed the source code and based on `packages/types/app/index.d.ts`, I inferred that the other two were `server` and `static`.


danielroe (approved):
Thank you ♥️

PR #1377, opened 2021-03-29
fix broken links

RomainMazB:
Hi there!

Reading the french documentation, I've found many broken links (+ not redirected), so I've made a search and replace into all the `/guides` directory and found many of these into other languages.


aytdm (approved):
Thank you :blush:

---
I found the redirect setting in `_redirects`. I wonder why it doesn't work properly.

https://github.com/nuxt/nuxtjs.org/blob/1171eb275550655cc6ffcd78d39b9505f6bc4d16/_redirects#L7

danielroe (approved):
Thank you for this ❤️

PR #1391, opened 2021-04-06
Add example explanation for transpile

iwt-nduesing:
Explaining what the example means also explains what `isModern` and `isLegacy` means.

### DIFF
--- a/content/en/guides/configuration-glossary/configuration-build.md
+++ b/content/en/guides/configuration-glossary/configuration-build.md
@@ -747,7 +747,7 @@ Starting with `v2.9.0`, you can also use a function to conditionally transpile, 
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


content/en/guides/configuration-glossary/configuration-build.md
```diff
@@ -747,6 +747,7 @@ Starting with `v2.9.0`, you can also use a function to conditionally transpile,
   }
 }
 ```
+With this example the dependency `ky` will be transpiled by Babel if nuxed is not in [modern mode](/docs/2.x/configuration-glossary/configuration-modern).
```
danielroe, line 750:
```suggestion
In this example, `ky` will be transpiled by Babel if Nuxt is not in [modern mode](/docs/2.x/configuration-glossary/configuration-modern).
```

danielroe (approved):

PR #1414, opened 2021-05-01
Update heroku-deployment.md

valentinoli:
* You do not need to declare a Procfile if the `start` script runs `nuxt start`.
* `NODE_ENV=production` by default on Heroku


Atinux:
Nice, normally when creating a nuxt app, the start command should be nuxt start in the package.json

I am thinking that we should instead explain to make sure they configure the start command of their package.json instead to even mention the Procfile. Linking for more advanced use cases to Heroku docs

valentinoli:
I agree it is better not to mention the Procfile at all. In addition, it should be mentioned that you need to add the Node.js buildpack to the Heroku app.

valentinoli:
I updated the PR.

* Removed Procfile related docs
* Removed NODE_ENV config since it is production by default
* Mentioned everything can be done either via Heroku dashboard or Heroku CLI
* Added how to create an app via CLI
* Added how to add Node.js buildpack
* Updated image of Settings in the dashboard
* Mentioned automatic deployment option

Atinux (approved):

danielroe (approved):

PR #1431, opened 2021-05-11
docs: solve confusing on `watchQuery` description

nikolausliu:
The new `fetch` hook introduced in 2.12 is not affected by `watchQuery`，but the description cause a confusion. modify it to match this chapter: https://nuxtjs.org/docs/2.x/components-glossary/pages-watchquery


Atinux:
Thanks @nikolausliu

PR #1451, opened 2021-05-20
Deploy to Koyeb Serverless Platform guide

edouardb:
Add Koyeb as a deployment provider option.


content/en/guides/deployment/deployment-koyeb.md
```diff
@@ -0,0 +1,133 @@
...
+## Dockerize your application
+
+To Dockerize your Nuxt application, you need to create a `Dockerfile` in your project directory containing the content below:
+
+```dockerfile
+FROM node:lts
+
+WORKDIR /app
+
+COPY . ./
```
pi0, line 55: Single step build will include all dev/build dependencies to final image. Can we update it with a Dockerfile like [this](https://github.com/nuxt/nuxt.js/blob/dev/examples/docker-build/Dockerfile)?

content/en/guides/deployment/deployment-koyeb.md
```diff
@@ -0,0 +1,133 @@
...
+
+```bash
+docker run -p 3000:3000 <YOUR_DOCKER_HUB_USERNAME>/my-nuxt-project
+```
+
+Open your browser and navigate to http://localhost:3000 to view your project landing page.
+
+## Push your Docker image on the Docker hub
+
+Since our Docker image is built and functional in our test, we can now upload it to the Docker Hub. In your terminal run the command below to push the image:
```
pi0, line 104: Maybe worth mentioning alternative container registries like [github container registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry. Also any hope users can benefit from continues-deployments? (we can have a call if you would like to check possibilities)

content/en/guides/deployment/deployment-koyeb.md
```diff
@@ -0,0 +1,133 @@
...
+Check the box `Use a private registry` and, in the select field, click **Create Registry Secret**.
+
+A modal opens asking for:
+
+- a name for the Secret which will be created, we can use for instance `docker-hub-secret`
+- the registry provider to generate the secret containing your private registry credentials, in our case Docker Hub
+- your Docker Hub username and password. We recommend you to [generate an access token](https://hub.docker.com/settings/security) from the Docker Hub to use instead of your password.
+  Once you've filled all the fields, click the **Create** button.
+
+In the _Ports_ section, change the export port from `80` to `3000`, which is the port our Nuxt application is listening on. This setting is required to let Koyeb know which port your application is listening to and properly route incoming HTTP requests. We don't need to change the _Path_, our app will be available at the root of our domain: `/`.
```
pi0, line 125: If the default port for Koyeb is `80` (and `PORT` environment variable is not available), we may include it in the docker image to reduce user steps for integrations. (also i _guess_ path can be somehow mentioned for [router.base](https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-router/#base))

pi0 (commented):
Thanks! I've put few comments for improvement.

content/en/guides/deployment/deployment-koyeb.md
edouardb: Will update to use multi-stage build.

content/en/guides/deployment/deployment-koyeb.md
edouardb: Of course, do you prefer the guide to use ghcr instead of Docker Hub, or mention the list of alternatives is enough? Users can benefit from CD by triggering a Koyeb deployment in their CD workflows. We are working to ease this part, in the coming weeks we will provide support of git-driven deployment which will significantly ease the flow to deploy Nuxt apps on Koyeb.

pi0: Just mentioning possibilities would be enough (and maybe also spoiler on upcoming changes? 🔥 )

content/en/guides/deployment/deployment-koyeb.md
edouardb: Will fix the port to remove this step. About the `path` parameter the user has to set, it's a Koyeb related setting which indicates our global load balancer where to forward requests. i.e. if your router base path is set to`/apps` and  you set the Koyeb path parameter to `/`, the app will be accessible via `/apps`. if the Koyeb path is set to `/nuxt` the app will be accessible via `/nuxt/apps.

content/en/guides/deployment/deployment-koyeb.md
edouardb: Spoiler currently hidden in the intro 🙃 `The platform lets you seamlessly run Docker containers, web apps, and APIs with git-based deployment`

edouardb:
@pi0 does it look ok for you with the latest edits?

pi0 (approved):

PR #1453, opened 2021-05-22
style: tiny rephrase in docs / store example

definiteIymaybe:
I believe 
> **add** the authenticated user to to our store

 would be more accurate here.


danielroe (approved):
Thanks - agreed. 👍

Issue #1457, opened 2021-05-25
Mobile view of blog page is overflown


shahbaz17:
I was reading https://nuxtjs.org/blog/creating-blog-with-nuxt-content on my mobile and saw the ```:src="require(`~/assets/images/${article.img}`)"``` is causing the overflow. Can someone confirm this? 

![Nuxt Blog code overflow](https://user-images.githubusercontent.com/6962565/119456679-8c16ac80-bd58-11eb-94dc-2d21a3cc29be.png)

This could be fixed by adding `word-break: break-word;` or `word-break: break-all;` to https://github.com/nuxt/nuxtjs.org/blob/main/assets/css/content.scss#L291

I would be happy to fix this.

shahbaz17:
Thanks @danielroe 

**This one is from Safari on iPhone 12**👇 
![NuxtorgSafarioniPhone12](https://user-images.githubusercontent.com/6962565/119498914-c8133700-bd83-11eb-88cf-73e339036eca.PNG)
**This one is from Chrome on iPhone 12** 👇 
![Nuxt orgChromeoniPhone12](https://user-images.githubusercontent.com/6962565/119498924-ccd7eb00-bd83-11eb-8e0e-04a6e02543d4.jpeg)

Steps to reproduce it:
* Visit https://nuxtjs.org/blog/creating-blog-with-nuxt-content on mobile browser
* Use two fingers to pinch the screen.

Here's a video to reproduce it. https://youtu.be/gVbTZXRSiG0

shahbaz17:
PR is ready for review @danielroe

PR #1459, opened 2021-05-25
Add word-break: break-word; to fix code overflow on markdown pages

shahbaz17:
This PR fixes https://github.com/nuxt/nuxtjs.org/issues/1457


Atinux:
Thanks ❤️

PR #1461, opened 2021-05-26
Title paramater should not be translated

Mamboleoo:
When checking the docs I found this example in French for the head function.
But I noticed the 'title' parameter was translated in French to 'titre' making the example invalid.


Atinux:
Thanks @Mamboleoo

PR #1467, opened 2021-06-01
docs: change minimum node version to 12

rchl:
According to v2.15.0 release notes [1] the minimum node version is 12.

[1] https://github.com/nuxt/nuxt.js/releases/tag/v2.15.0

### DIFF
--- a/content/en/guides/get-started/installation.md
+++ b/content/en/guides/get-started/installation.md
@@ -12,7 +12,7 @@ ManualInstallVideoTitle: Nuxt Manual Installation
 
 ## Prerequisites
 
-- [node](https://nodejs.org) - _We recommend you have the latest LTS version installed_ or at least v10.13 and above (excluding v11).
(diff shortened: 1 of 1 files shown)


content/en/guides/get-started/installation.md
```diff
@@ -12,7 +12,7 @@ ManualInstallVideoTitle: Nuxt Manual Installation
 
 ## Prerequisites
 
-- [node](https://nodejs.org) - _We recommend you have the latest LTS version installed_ or at least v10.13 and above (excluding v11).
+- [node](https://nodejs.org) - _We recommend you have the latest LTS version installed_ or at least v12 and above.
```
danielroe, line 15:
```suggestion
- [node](https://nodejs.org) - _We recommend you have the latest LTS version installed_.
```

rchl: OK. That will remove the info about minimum supported version (even if it's not exactly phrased that way). I guess that's better from maintenance point of view.

danielroe (approved):

PR #1468, opened 2021-06-01
content: replace blog articles covers

clemcode:


R-mooon:
It is specified on official unsplash plugin that it is not a necessity
image

R-mooon:
Okay, then i will retrieve them :)

clemcode:
I agree with the credits, I'll handle the credits placeholder and we'll find every author

PR #1470, opened 2021-06-02
docs: update `render.static` false option

danielroe:
**Will need to wait until release of 2.16**

context: https://github.com/nuxt/nuxt.js/pull/9365


Atinux:
@danielroe do you mind updating it to the correct branch>

Atinux (approved):

PR #1480, opened 2021-06-11
docs: add ben to ambassador list

bencodezen:
It seems like there's some auto formatting happening for the CSS styles that I'm not committing myself. Hope this is alright!


clemcode (approved):
👍

PR #1496, opened 2021-06-24
Add note about script lang for typescript JSX

Epskampie:
This tripped me up for a long time, hopefully you will accept this small addition to the docs.


edwardz8:
hope this makes it in! nuxt + vue 3 + jsx or tsx would be too cool!

Epskampie:
> hope this makes it in! nuxt + vue 3 + jsx or tsx would be too cool!

It’s already working in current nuxt. Only the fact that fragments are missing in vue 2 makes jsx kind of painful.

Atinux (approved):

PR #1523, opened 2021-07-10
feat(preview): lazy load more homepage images

btkostner:
Adds `loading="lazy"` to the "Sharing is Caring" and "Testimonials" sections on the homepage.


btkostner:
Ahhhh whoops. Sorry about the delay there. The notification must have been marked done on accident 😬

PR #1524, opened 2021-07-10
feat(preview): add some accessability props to homepage elements

btkostner:
This adds some accessibility warnings caught from lighthouse.

I think the `docus` `AppAside` organism could use a little love too but I'm not going to talk through the whole thing here.

### DIFF
--- a/components/organisms/animations/AnimFromCliCodeblock.vue
+++ b/components/organisms/animations/AnimFromCliCodeblock.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="fromcli__container">
+  <div class="fromcli__container" aria-live="polite">
     <div class="fromcli__code">
       <div class="fromcli__code__wrapper">
         <div class="fromcli__code__bash">
@@ -114,23 +114,25 @@
     </div>
     <div class="fromcli__display">
       <div class="fromcli__display_loading">
-        <img loading="lazy" class="fromcli__display__logo" src="~/static/img/home/learn/master/logo.svg" />
+        <img loading="lazy" class="fromcli__display__logo" src="~/static/img/home/learn/master/logo.svg" alt="" />
         <div class="fromcli__display__progressbar">
           <div class="fromcli__display__progressbar__background"></div>
           <div class="fromcli__display__progressbar__progress"></div>
         </div>
       </div>
       <div class="fromcli__display__content">
-        <img loading="lazy" class="fromcli__display__content__logo" src="~/static/img/home/learn/master/logo.svg" />
+        <img loading="lazy" class="fromcli__display__content__logo" src="~/static/img/home/learn/master/logo.svg" alt="" />
         <img
           loading="lazy"
           class="fromcli__display__content__content1"
           src="~/static/img/home/learn/master/cliContent1.svg"
+          alt=""
         />
         <img
           loading="lazy"
           class="fromcli__display__content__content2"
           src="~/static/img/home/learn/master/cliContent2.svg"
+          alt=""
         />
       </div>
     </div>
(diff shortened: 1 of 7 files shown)


components/organisms/animations/AnimFromScratchCodeblock.vue
```diff
@@ -57,16 +57,16 @@
               <span>&lt;</span>
               <span class="text-red-400">h1</span>
               <span>&gt;</span>
-              Hello world !
+              <span>Hello world !</span>
```
btkostner, line 60: This fixes some extra spacing in the animation:
image

components/organisms/app/AppFooter.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <footer class="" aria-labelledby="footerHeading">
```
btkostner, line 2: There is no `footerHeading` id on the page, and also not really needed.

components/organisms/animations/AnimFromCliCodeblock.vue
```diff
@@ -114,23 +114,25 @@
     </div>
     <div class="fromcli__display">
       <div class="fromcli__display_loading">
-        <img loading="lazy" class="fromcli__display__logo" src="~/static/img/home/learn/master/logo.svg" />
+        <img loading="lazy" class="fromcli__display__logo" src="~/static/img/home/learn/master/logo.svg" alt="" />
```
manniL, line 117: We possibly should set `role="presentation"` as well here ☺️

components/organisms/animations/AnimFromCliCodeblock.vue
```diff
@@ -114,23 +114,25 @@
...
         <div class="fromcli__display__progressbar">
           <div class="fromcli__display__progressbar__background"></div>
           <div class="fromcli__display__progressbar__progress"></div>
         </div>
       </div>
       <div class="fromcli__display__content">
-        <img loading="lazy" class="fromcli__display__content__logo" src="~/static/img/home/learn/master/logo.svg" />
+        <img loading="lazy" class="fromcli__display__content__logo" src="~/static/img/home/learn/master/logo.svg" alt="" />
```
manniL, line 124: See above

components/organisms/home/HomeTestimonials.vue
```diff
@@ -84,16 +85,18 @@
                   <NuxtLabel tag="span" class="font-bold text-base">{{ testimonial.author }}</NuxtLabel>
                   <NuxtLabel tag="span" class="text-sm dark:text-cloud-lighter">{{ testimonial.job }}</NuxtLabel>
                 </a>
-                <a :href="testimonial.jobUrl" target="_blank" rel="noopener sponsored" class="hidden xl:block">
+                <a :href="testimonial.jobUrl" :aria-label="`${testimonial.jobName} website`" target="_blank" rel="noopener sponsored" class="hidden xl:block">
```
manniL, line 88: Is there a reason you prefer `aria-label` over `title` here?

Tahul:
Hey @btkostner ; thank you a lot for the PR 😄

Would love to get your insights about AppAside!

@bdrtsky ; could you take a look at this PR?

components/organisms/home/HomeTestimonials.vue
btkostner: No reason. I can switch it to `title` instead. I didn't check to see if there was a convention for it in the project.

components/organisms/animations/AnimFromCliCodeblock.vue
btkostner: I think we can place it on `div.fromcli__display` so it applies to all of the child images as well.

btkostner: I applied `role="img"` to both of the animation components one level up. I'm not sure if we need to set the role on this specifically. Or if we should change it to `presentation`.

btkostner:
@Tahul Because it's a slide out navigation, I'd include on the menu buttons (and the back button):

```html
aria-label="open navigation" aria-controls="id of the aside" :aria-expanded="$menu.visible"
```

I'd also add something like `esc` keybind to close the navigation menu.

PR #1543, opened 2021-07-16
fix: remove localPath to external links

Flosciante:
related to [https://github.com/nuxtlabs/nuxtjs.org-board/issues/136](url)


clemcode (approved):
👍

PR #1546, opened 2021-07-16
feat: sponsors refactoring

clemcode:
This PR goal is to make a single source of truth for sponsors and dedup them by extracting folder to root. (/content/_sponsors)


Atinux (commented):
Looks great. One note, I would slugify the `tier` and use i18n for displaying them as text so we can translate them

.gitignore
```diff
@@ -8,4 +8,5 @@ coverage
 dist
 sw.*
 .env
-.output
\ No newline at end of file
+.output
+.husky
```
Flosciante, line 12: 🐕

Flosciante (approved):
👍

PR #1579, opened 2021-07-25
docs: add definitions for acronyms + technical terms

samilamti:
This PR attempts to fix things that caught my attention while reading this fine article, namely:
- Explains what HMR is. 
- Removes some redundant words. 
- Cross-references the store directory when referencing Vuex.
- Links out to the MDN glossary definition of Falsy.
- Explains what DX/UX is.

It'd be great to have a acronym/definition type component that one could use, e.g. 

```html
<acronym def="Hot Module Reload / Replacement">HMR</acronym>
```

styled e.g. as

```css
display: inline-block;
border-bottom: dashed 1px silver;
cursor: help;
```

### DIFF
--- a/content/en/docs/2.concepts/2.context-helpers.md
+++ b/content/en/docs/2.concepts/2.context-helpers.md
@@ -15,7 +15,7 @@ The context provides *additional* and often optional information about the curre
 
 The `context` object is available in specific Nuxt functions like [asyncData](/docs/features/data-fetching#async-data), [plugins](/docs/directory-structure/plugins), [middleware](/docs/directory-structure/middleware) and [nuxtServerInit](/docs/directory-structure/store#the-nuxtserverinit-action). It provides _additional_ and often optional information about the current request to the application.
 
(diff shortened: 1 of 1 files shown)


content/en/guides/concepts/context-helpers.md
```diff
@@ -132,11 +132,11 @@ export default {
 }
 ```
 
-Want to use query parameters instead? You then use [context.query.id](/docs/2.x/internals-glossary/context#query) then.
+Want to use query parameters instead? You then use [context.query.id](/docs/2.x/internals-glossary/context#query).
 
 ### Redirecting users & accessing the store
 
-Accessing the Vuex store (when you have it set up through the `store` directory) is also possible through the context. It provides a `store` object which can be treated as `this.$store` in Vue components. In addition, we use the `redirect` method, an exposed helper through the context, to redirect the user in case the `authenticated` state is falsy.
+Accessing the [Vuex store](https://nuxtjs.org/docs/2.x/directory-structure/store) (when you have it set up through the `store` directory) is also possible through the context. It provides a `store` object which can be treated as `this.$store` in Vue components. In addition, we use the `redirect` method, an exposed helper through the context, to redirect the user in case the `authenticated` state is [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy).
```
danielroe, line 139:
```suggestion
Accessing the [Vuex store](/docs/2.x/directory-structure/store) (when you have it set up through the `store` directory) is also possible through the context. It provides a `store` object which can be treated as `this.$store` in Vue components. In addition, we use the `redirect` method, an exposed helper through the context, to redirect the user in case the `authenticated` state is [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy).
```

Atinux (dismissed):

Atinux:
Do you mind applying these changes to the `preview` branch too @samilamti 🙏 ?

danielroe (approved):
thanks ❤️

PR #1582, opened 2021-07-26
docs: change example to `head` object and remove resource hints

samilamti:
This PR attempts to clarify the difference between setting the `head` to an object vs. as a function. It also removes _Resource Hints_ as they belong to the render configuration, not metadata and SEO https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-render#resourcehints


Atinux (dismissed):

danielroe (approved):

pi0:
I think we should rebase PR for `main` branch?

PR #1596, opened 2021-07-29
docs(ja): ja translation for preview

kazupon:


kazupon:
@Atinux 
I've just translated `v3.md`

PR #1597, opened 2021-07-29
Update configuration.md for host property.

seandan7:
Update to note that protocols should not be set when setting `server.host` property

### DIFF
--- a/content/en/guides/features/configuration.md
+++ b/content/en/guides/features/configuration.md
@@ -467,6 +467,12 @@ export default {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Atinux:
Why is that?

seandan7:
I was getting an error
```Error: getaddrinfo ENOTFOUND```
 when including it, and was unable to come across a solution until finding this. 
https://stackoverflow.com/a/28385129/7044030

I just thought it could be helpful for anyone going through the docs and changing to a production host.

PR #1599, opened 2021-08-01
docs: correct name of component interacting with router-middleware

crookedneighbor:
Just a minor thing I noticed. The example code does not have a `components/Navigation.vue` file, and instead, the `layouts/default.vue` file is what is using the value of the class from the store.


inouetakuya (approved):
Thank you for your contribution :+1:

danielroe (approved):

danielroe:
Thanks!

PR #1618, opened 2021-08-11
fix: modules page

Tahul:
Here is a work in progress fix for the modules page, taking the logic out of the components.

It needs a review from @danielroe concerning the few errors remaining, seem related to usage of Compositon API, pretty sure I'm wrong somewhere but after a lot of investigation I would love a little help.

I also added the linting back to the project, and fixed all the errors on static generation related to image paths and 404 links. 😄


Atinux:
Looks great! But seems the `Sort By` select is not working (cc @Tahul @Flosciante )

PR #1619, opened 2021-08-11
docs: add options usage to example

escapedcat:
I was searching forever to see how to use `options` in a plugin. Is this mentioned in the docs somewhere? Couldn't find it.  

Looking at these helped:
- https://stackoverflow.com/questions/57756064/pass-data-from-nuxt-module-to-plugin
- https://github.com/nuxt-community/community-modules/issues/170#issuecomment-346764312

Happy to close this if this is documented somewhere. Happy to modify this as well to make this fit better into the current docs.

### DIFF
--- a/content/en/guides/directory-structure/plugins.md
+++ b/content/en/guides/directory-structure/plugins.md
@@ -259,6 +259,8 @@ Sometimes you want to make functions or values available across your app. You ca
 
 Nuxt.js provides you with an `inject(key, value)` method to do this easily. Inject is given as the second parameter when exporting a function. The `$` will be prepended automatically to the key.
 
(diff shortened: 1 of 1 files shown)


Kronhyx (approved):

escapedcat:
@Atinux thanks, that's actually the info I would have needed! So it's basically already documented but I didn't see the connection.  

People like me might help a link from from the plugins-section to modules-section where this is mentioned.  
Not sure if this is wanted. Happy to adjust this.

danielroe (approved):

Atinux:
Thank you, it would be awesome if you could also apply the changes to the `preview` branch 😊

PR #1622, opened 2021-08-13
docs: add note on modules using plugins

hannakim91:
Nuxt.js behavior is different when resolving modules vs modules with plugins. A note for clarification might be helpful. 

Wondering if the [modules directory page](https://nuxtjs.org/docs/2.x/directory-structure/modules) can be updated too?


Atinux:
Thank you @hannakim91 

Can you please update it in the `preview` branch?

PR #1639, opened 2021-08-25
docs: add note on modules using plugins

hannakim91:
See: https://github.com/nuxt/nuxtjs.org/pull/1622


Atinux:
Thank you @hannakim91 💚

PR #1644, opened 2021-08-28
docs(ja): translate example for preview

kazupon:


kazupon:
I've just finished `1.examples` docs translation!
ping @Atinux

PR #1677, opened 2021-09-13
Feat/content/404

R-mooon:


pi0:
We need to rewrite `preview`  branch history in order to:
- Preserve old contributors to the same content
- Remove (big) design files and move them to `nuxt/design` 
- Avoid mixed content and separate `content/`  changes

PR #1698, opened 2021-09-15
fix: broken link

debs-obrien:
fix link to video courses


Atinux:
Thanks @debs-obrien 💚

PR #1721, opened 2021-09-16
feat(accessibility): make focus rings green & only visible on tab focus

smarroufin:
Resolves nuxtlabs/nuxtjs.org-board#180


Flosciante (approved):
👍

clemcode (approved):
👍

PR #1745, opened 2021-09-20
docs(ja): update 1.fetch.md

Yuki-Inoue:
Currently, the purpose of `fetchDelay` is explained as "過剰実行を防ぐため", which means "to avoid excessive execution".
I think this is not correct.

I propose changing the phrase to "極短期間での画面切り替えによるちらつきを防ぐため", which is google translated as:
"To prevent flicker due to screen switching in a very short period of time"

(Note: the original phrase was: "to avoid quick flashes", which I could not come up with concise phrase, so a bit descriptive)


kazupon (commented):
It's good, so it's a clear translation and includes context. :)
Thank you for your feedback!

PR #1748, opened 2021-09-20
refactor: heroes unified & slots-compliant

smarroufin:
- Removed unused heroes
- Used `PageHero` as much as possible (except Home & Nuxt3 heroes)
- Transformed all content props as slots


Atinux:
Nice work @smarroufin 
Next step would be to refactor more components to be more generic (such as not having to need `PartnersBottomHero.vue` for example)

PR #1753, opened 2021-09-21
feat(events): improve UI

smarroufin:
Resolves nuxtlabs/nuxtjs.org-board#224


danielroe (approved):

manniL:
Looks great! Only one issue: "Watch video" is not clickable (and does not link to the video)

grafik

PR #1770, opened 2021-09-22
feat: cookie banner

Flosciante:


components/theme/app/AppLayout.vue
```diff
@@ -15,8 +16,9 @@
         <slot />
       </div>
     </div>
+    <AppFooter :links="footerLinks" :class="showCookieBanner ? 'pb-16 md:pb-12' : ''" />
 
-    <AppFooter :links="footerLinks" />
+    <CookieBanner class="w-full fixed bottom-0 left-0 z-40" @cookie-banner="showCookieBanner = false" />
```
smarroufin, line 21: Really minor comment, but I feel like doing `inset-x-0` is more secure and simpler than `w-full left-0` 🙂

components/templates/CookieBanner.vue
```diff
@@ -0,0 +1,62 @@
...
+          </SectionButton>
+        </template>
+      </CookieConsent>
+    </VueIfBot>
+  </div>
+</template>
+
+<script>
+import VueIfBot from 'vue-if-bot/dist/vue-if-bot.es'
+import CookieConsent from 'vue-cookieconsent-component/src/components/CookieConsent.vue'
```
smarroufin, line 54: According to the docs it does not seem right to import them this way. Is there a reason?
https://github.com/Developmint/vue-if-bot#synchronous-import
https://github.com/EvodiaAut/vue-cookieconsent-component#usage

smarroufin (changes requested):
Out of thoses two comments, it seems fine and works well. Probably wait for a second review, from someone actually knowing cookies 😅 .

components/theme/app/AppLayout.vue
Flosciante: 👍

components/templates/CookieBanner.vue
manniL: I think I have to update the docs 🙈 
`vue-if-bot` import seems alright 👍 
https://github.com/manniL/lichter.io/blob/31cded247d5c2cad2bb6f29be5355e46bec9ffdb/layouts/default.vue#L18

PR #1772, opened 2021-09-22
refactor: release on datetime with named formatting

kazupon:
Using named formatting is much easier to maintain.
https://kazupon.github.io/vue-i18n/guide/formatting.html#named-formatting

The message of an i18n resource can understand context-sensitive by using parameters.
From a localization point of view, you can also benefit from named formatting.


Atinux (approved):
I let @smarroufin confirm, thank you @kazupon ❤️

smarroufin (approved):
I actually did this 😞  and I couldn't agree more with you. This is the best practice.

PR #1778, opened 2021-09-23
fix(redirects): add localised urls

smarroufin:
Ugly? Yes. Urgent? Yes.

I know it probably shouldn't be fixed this way. But it at least fixes all those Google searches leading to a 404. We will clean that in time.

Also added a missing error code 🤯  and removed commented rules not needed anymore.


danielroe (approved):

pi0:
Seems good until we fix `/2.x` issue on docus :)

PR #1781, opened 2021-09-24
docs: Include Page.key option in documentation

mikeapr4:
I stumbled across this option reading the code, I think it would be very worthwhile to document it.

It would also explain why route navigation in Nuxt will destroy/recreate page components even if the same component is used in two routes. I'm not debating the behaviour at all, I agree with it, but as it doesn't match the norm for a Vue project, no harm having something in the documentation which explains it.


danielroe (approved):
This looks good to me, and a good addition to the docs. Thank you ❤️

Atinux (approved):

PR #1787, opened 2021-09-26
fix: home hero title selectable

### LINKED ISSUE #1786: The homepage module level is incorrect, so the title cannot be copied
for example:
Visit: https://nuxtjs.org/

You need to adjust the z-index of this module to be relatively high, so that you can choose "The Intuitive Vue
Framework》
image

Geminii:
Resolves #1786

### DIFF
--- a/components/organisms/home/HomeHero.vue
+++ b/components/organisms/home/HomeHero.vue
@@ -5,7 +5,7 @@
     <div class="absolute left-0 w-full h-full overflow-hidden -top-24 z-0">
       <Star v-for="star in 80" :key="star" />
     </div>
-    <div class="flex flex-wrap justify-center py-0 section">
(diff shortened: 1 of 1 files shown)


components/organisms/home/HomeHero.vue
```diff
@@ -5,7 +5,7 @@
     <div class="absolute left-0 w-full h-full overflow-hidden -top-24 z-0">
       <Star v-for="star in 80" :key="star" />
     </div>
-    <div class="flex flex-wrap justify-center py-0 section">
+    <div class="flex flex-wrap justify-center py-0 section relative z-20">
```
smarroufin, line 8: This means you can remove the two `z-20` on the below `SectionButton`s.

smarroufin (changes requested):

PR #1807, opened 2021-09-28
feat: (HomeHero) get started button instead of npm button

Flosciante:


smarroufin (changes requested):
While you're at it, can you move the `z-20` like this: https://github.com/nuxt/nuxtjs.org/pull/1787/files
I'll then dismiss the other PR, to avoid conflicts and endless adjustments.
This will fix nuxt/nuxtjs.org#1786

smarroufin (approved):

PR #1816, opened 2021-10-01
feat: (header) new dropdown

Flosciante:


content/fr/_collections/navigations/header.md
```diff
@@ -4,82 +4,135 @@ links:
...
+        color: "bg-mint-lighter"
   -
-    title: 'Partenaires'
-    slug: 'partners'
-    to: '/partners'
+    title: "Partenaires"
+    slug: "partners"
+    to: "/partners"
+    subtitle: "subtitle"
+    icon: "icon"
```
smarroufin, line 137: I guess this was testing values and should be deleted

components/molecules/ItemHeaderDropdown.vue
```diff
@@ -0,0 +1,48 @@
+<template>
```
smarroufin, line 1: I would rename `ItemHeaderDropdown` to `HeaderDropdownItem`. It matches better our naming convention.

components/theme/app/AppHeader.vue
```diff
@@ -22,6 +22,12 @@
             :items="[link.items]"
             placement="bottom"
             mode="hover"
+            :dropdown-menu-class="
+              ('h-full', isHome ? 'bg-sky-darker text-white' : 'bg-cloud-surface dark:bg-sky-darker')
+            "
+            dropdown-class="w-max"
+            :open-delay="0"
+            :items-class="`py-1 grid grid-cols-${Math.round(link.items.length / 3)} gap-x-12`"
```
smarroufin, line 30: I'm not really sure but the computation seems wrong.
You should divide by 4 and you should `ceil` instead of `round`: `Math.ceil(link.items.length / 4)`
Meaning that [from 1 to 4] -> [1 col], [from 5 to 8] -> [2 cols], etc.
Or I did miss something

smarroufin (changes requested):

smarroufin (approved):

pi0:
Shall we also include new structure in PR? (https://github.com/nuxtlabs/nuxtjs.org-board/issues/222)



image

PR #1822, opened 2021-10-04
Fix Japanese translation

karur4n:
English text

```
If you have used vue-router before, you can think of the <NuxtLink> as a replacement for <RouterLink>
```


kazupon (approved):
Good catch!
Thanks! 💚

PR #1908, opened 2021-11-02
feat: disable mailto links in favor of clipboard copy

smarroufin:
In order to "disable" `mailto:` links (making emailing too massive, including spam), those links now simply copy the email address to the clipboard and display a toast.


plugins/clipboard.client.js
```diff
@@ -0,0 +1,30 @@
+import { useNotifications } from '~/plugins/notifications'
+
+export default ({ $t }, inject) => {
+  const { add: addNotification } = useNotifications()
+
+  inject('clipboard', {
+    copy: (text, successMsg, failureMsg) => {
+      if (!navigator.clipboard) {
```
pi0, line 8: Maybe a simple `location.replace` if not?

smarroufin: Thanks for the review. But I don't get it 😅 
What do you mean? There should not be any redirect 🤔

pi0: We can return `false` here or throw an unsupported error and handle it in `copyEmailToClipboard`. My point is, we are preventing the default behavior of link clicking and changing it to "copy to clipboard". While this can be solved by doing a programmatic navigation that bots cannot follow (same as [email obfuscation in cloudflare](https://support.cloudflare.com/hc/en-us/articles/200170016-What-is-Email-Address-Obfuscation-))

smarroufin: Oh yes, I see what you mean. Agreed, we should still do something even if Clipboard is not supported.
I like the simple plan of returning false and then let the `mailto` proceed. Updating it!

clemcode (changes requested):
I think we miss JA translation for the string

clemcode (approved):
👍

PR #1910, opened 2021-11-02
chore(env): fix env variables usage on runtime

smarroufin:
Fixes the usage of environnement variables on runtime. Should use `publicRuntimeConfig` instead of `process.env`.


clemcode (approved):
👍

PR #1912, opened 2021-11-02
chore: redirect `/modules` to `modules.nuxtjs.org`

smarroufin:
To merge even if https://modules.nuxtjs.org is not redesigned. At least it works well 🙂


Atinux:
I think we could also update the link from the header and footer.

clemcode (approved):
👍

Atinux (approved):

PR #1926, opened 2021-11-11
docs(ja): fix mini typo 6.internals-glossary/9.nuxt.md

karuhi:
There was a very small typo, which has been corrected.


kazupon (approved):
Good catch!
Thanks!

PR #1945, opened 2021-11-20
[doc] $http is not directly accessible from context

Wolvborg:
Changed $http to app.$http

### DIFF
--- a/content/en/docs/2.concepts/2.context-helpers.md
+++ b/content/en/docs/2.concepts/2.context-helpers.md
@@ -81,12 +81,12 @@ With [ES6](https://hacks.mozilla.org/2015/05/es6-in-depth-destructuring/) you ca
 
 ```js{}[pages/posts/_id.vue]
 export default {
-  async asyncData({ params, $http, error }) {
+  async asyncData({ params, app, error }) {
     const id = params.id
 
     try {
       // Using the nuxtjs/http module here exposed via context.app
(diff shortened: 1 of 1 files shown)


Wolvborg:
Will be my first contribution for an open source :p(though very tiny)

Wolvborg:
@Atinux  can you please approve

danielroe (commented):
@Wolvborg What makes you think that `$http` isn't directly accessible from context?

https://codesandbox.io/s/blissful-darkness-5r2m5?file=/pages/index.vue

If it's not showing up on the types, then would you raise an issue [here](https://github.com/nuxt/http/issues) instead?

Wolvborg:
@danielroe It is written in the docs that context consist followings keys

image


Also in the same page(https://nuxtjs.org/docs/concepts/context-helpers), we are accessing `$http` from `app`, so how we can destructure `$http` from `context`. 
image


This is typo error in docs and not in the actual code 
image

danielroe:
@Wolvborg It is accessible through both `context.$http` and `context.app.$http`. Here is the [documentation for the http module](https://http.nuxtjs.org/getting-started/usage#using-in-asyncdata).

If it's helpful, here's [some information](https://nuxtjs.org/docs/directory-structure/plugins/#inject-in-root--context) on what 'inject' does if you are injecting your own values into context.

PR #1950, opened 2021-11-25
feat: update sponsors (vercel and qoddi) with collection refactor

clemcode:
- Lift `_collections/sponsors` at the root level of `content` folder to ease maintainance, as we don't need to localize them
- Add Vercel as MVP partner
- Update Flashdrive sponsor following their rebranding


smarroufin (changes requested):
- You need to remove the old flashdrive png since you added a new one for qoddi
- Please make sure you properly deleted all `/content/en/_collections/sponsors` files, it does not seem to be the case

smarroufin (approved):

PR #1951, opened 2021-11-25
fix: BannerMasteringNuxt image overflow

anandsiddharth:
fixed banner's background image overflow

Screenshot 2021-11-26 at 12 01 20 AM


smarroufin:
Hi @anandsiddharth 
Nice catch et thanks for the fix! :green_heart:

smarroufin (approved):

PR #1952, opened 2021-11-26
Feat improve partner contact

smarroufin:
- Clears form fields on success
- Makes all fields required
- Adds Textarea component (fixes font, adds auto-resize)


clemcode (approved):
👍

PR #1959, opened 2021-12-01
feat(docus): upgrade Docus to v2 / prepare Nuxt 3

Tahul:
Hey 👋 ;

This PR updates the nuxtjs.org website with all latest changes from `docus` upgrades.

This also prepares the repository to be migrated to {Nuxt/Docus}3 once it will be released.

This PR removes any usage from `@nuxtjs/composition-api` and all its features, and replaces it with latest version of `@nuxt/bridge`.

It is still WIP and has been worked with @smarroufin. 😄

Any comments are welcome!

cc @Atinux @Flosciante @pi0 @clemcode


components/atoms/links/AppLink.vue
```diff
@@ -27,7 +27,7 @@ export default defineComponent({
     }
   },
   setup(props) {
-    const { i18n, $clipboard } = useContext()
+    const { i18n, $clipboard } = useNuxtApp().vue2App
```
danielroe, line 30: We should be able to access these directly from `useNuxtApp()` now.

PR #1961, opened 2021-12-03
Fix typo in 2.assets.md

rikoroku:


danielroe (commented):
I know it looks like it might be a typo, but Stylus is a CSS preprocessor: https://stylus-lang.com/

PR #1970, opened 2021-12-30
feat: add location to partners list cards

clemcode:
Capture d’écran 2021-12-30 à 10 05 26
Capture d’écran 2021-12-30 à 10 05 43


smarroufin (approved):
LGTM

PR #1976, opened 2022-01-08
docs: add template wrapper for client-only placeholder

KareemDa:
I've tried to add `#placeholder` in my placeholder component inside `client-only` but it didn't work.
It worked after changing it into `slot="placeholder"`

### DIFF
--- a/content/en/docs/3.features/8.nuxt-components.md
+++ b/content/en/docs/3.features/8.nuxt-components.md
@@ -328,7 +328,9 @@ Use a slot as placeholder until `<client-only />` is mounted on client-side.
       <comments />
 
       <!-- loading indicator, rendered on server-side -->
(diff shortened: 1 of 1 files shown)


danielroe:
This syntax is now deprecated: https://vuejs.org/v2/guide/components-slots.html#Named-Slots-with-the-slot-Attribute

This explains the current syntax, add used in the docs: https://vuejs.org/v2/guide/components-slots.html#Named-Slots-Shorthand

I suspect you are using an older version of Vue.

KareemDa:
Thank you for your response, Actually, I'm using `2.6.14` ,  that's right I've added an old syntax. 
but also shouldn't we wrap the component with template ?? 

I'm getting lint error 
image

PR #1978, opened 2022-01-09
feat: add component watchParam

### LINKED ISSUE #1977: [Doc] `watchParam` component property
No one word about `watchParam` feature from this PR nuxt/nuxt.js#6244, in all languages

This caused issue nuxt/nuxt.js#10139

Kolobok12309:
fix #1977 
image
image


Uzere:
As I spent a lot of time figuring out what happens in nuxt/nuxt.js#10139, I would like to see watchParam menitioned on https://nuxtjs.org/docs/components-glossary/key#the-key-property because that is how one can get watchParam effect.
Also I think it would be good idea to explicitly menition when data() is called in https://nuxtjs.org/docs/concepts/nuxt-lifecycle#nuxt-lifecycle

Kolobok12309:
@Uzere, Actually agree with you, unfortunately, I have no experience in graphic design and will not be able to update hooks

PR #1983, opened 2022-01-12
fix: z-index issue Discover Partners/Explore section in the home page

ozgurg:
Hi. I realize Discover Partners section in the home page of nuxtjs.org has z-index issue on certain -especially high- resolutions. 

I just add "z-20" class to section and it is fixed.

Addition: I just realized that the same problem exists in the Explore section. :)

https://user-images.githubusercontent.com/6717356/149040131-5b8606ec-3879-49a5-b36a-7797d6d7b8f5.mov

### DIFF
--- a/components/organisms/home/HomeDiscoverPartners.vue
+++ b/components/organisms/home/HomeDiscoverPartners.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="relative pb-48 sm:pb-56 md:pb-64 lg:pb-72 xl:pb-92 2xl:pb-128 bg-sky-surface">
-    <section class="relative pt-20 z-20">
+    <section class="relative pt-20 z-10">
       <NuxtContainer class="flex flex-col items-center xl:pt-10 text-sky-black">
         <div class="flex flex-col w-full items-center col-span-12">
           <div class="mb-2">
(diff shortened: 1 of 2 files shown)


smarroufin:
Hi @ozgurg and thanks for the report!
It seems not many people with 4k screen have tested this page :sweat_smile:

components/organisms/home/HomeDiscoverPartners.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="relative pb-48 sm:pb-56 md:pb-64 lg:pb-72 xl:pb-92 2xl:pb-128 bg-sky-surface">
-    <section class="relative pt-20">
+    <section class="relative pt-20 z-20">
```
smarroufin, line 3: You could reduce it to `z-10` and remove `z-10` from the below `img`. It would simplify things and make sure z-indexes don't escalde too quickly.

smarroufin (changes requested):
If you don't mind, you can also apply this logic to `HomeExplore` (remove z-index from `img`s and add it to `section`)

components/organisms/home/HomeDiscoverPartners.vue
ozgurg: This is definitely a better solution! :)

ozgurg (commented):
This is definitely a better solution! :)

smarroufin (approved):

PR #1997, opened 2022-01-25
docs(fr): translate store directory structure

roiLeo:
- missing translation on directory-structure/store


smarroufin (approved):
Nice catch. Thanks! 💚

PR #2010, opened 2022-02-06
Docs/missing persian translation

soroushCoder:
The Persian (fa) content is already available, starting with the directory structure.


soroushCoder:
There were some problems that I prefer to close it

Issue #1987, opened 2022-01-14
[Translation] Guide Translation to Portuguese Brasil (pt-br)


felipesuri:
This question will be used to

a. Organize the process of translating the entire website to `pt-br`.
b. To gather all the information about the status of the translation of the whole website to `pt-br`.
c. Track all-volunteer translators (to know which part of the website they are translating, to avoid conflicts during the whole process).

If you would like to volunteer, you can comment below on which page you will be working on the translation, so that everyone knows.

Note: The main folder is /content/pt-br/

---

The docs pages translated:

- [x] Index (including Header and Footer).
- [x] docs/1.get-started
- [ ] docs/2.concepts (partly, only the page 1).
- [ ] docs/3.features
- [ ] docs/4.directory-structure
- [ ] docs/5.configuration-glossary
- [ ] docs/6.internals-glossary
- [ ] docs/7.components-glossary

I would like to ask @smarroufin or someone who is maintaining this part of the project for this question.

felipesuri:
@smarroufin Please, Can you add labels to this issue?

Like:
- Portuguese
- Translations

Please!

trizotti:
You guys are doing an amazing job translating to pt-BR! Can I help somehow? 😄

felipesuri:
Welcome, @trizotti. Actually I'm translating the docs/3.features. You can help me with this? This section is so big.

However, in case you don't want to help me with docs/3.features, you can translate any other section, I only ask to you say what are you doing.

trizotti:
Sure I can help with the 'docs' page! @felipesuri , any suggestion on how I could help with the 'docs' page? Is there any specific section you'd like to recommend for some parallel work?

trizotti:
## Examples
I'm currently working on this page translations, and as it gets done I'll update the following list:
- [x] 1.routing
- [x] 2.data-fetching
- [x] 3.assets-management
- [x] 4.transitions
- [x] 5.seo
- [x] 6.loading
- [x] 7.miscellaneous
- [ ] 8.middlewares
- [ ] 9.plugins
- [ ] 10.modules

felipesuri:
I don't understand the 'Examples' above, but, the pages on `docs/3.features` is:

- [x] 1.rendering-modes
- [x] 2.deployment-targets
- [x] 3.file-system-routing
- [ ] 4.data-fetching *WIP*
- [ ] 5.meta-tags-seo `*`
- [ ] 6.configuration `*`
[...]

PR #2019, opened 2022-02-22
Add pt-BR support for the Examples page

trizotti:
Continuing Portuguese Brazil translation of the docs. (https://github.com/nuxt/nuxtjs.org/issues/1987)


smarroufin (approved):
Hi @trizotti 
That is a lot of work, thanks for the contribution! 💚 

LGTM! Just waiting for @ChristopheCVB approval and we are good to go.

Just so you know, and you probably noticed, PT/BR are not yet available on prod. Donc worry this will be live at some point, there are stuff that need to be taken care of at NuxtLabs before we put it live.

ChristopheCVB (approved):
Amazing work! Thanks!

PR #2021, opened 2022-02-27
docs: clarify what 'this function' means in memory leak warning

literakl:
For me as a Nuxt newbie the original sentence has been confusing and I thougth it was in a conflict with previous chapters.

### DIFF
--- a/content/en/docs/4.directory-structure/10.plugins.md
+++ b/content/en/docs/4.directory-structure/10.plugins.md
@@ -231,7 +231,7 @@ export const actions = {
 ```
 
 ::alert{type="warning"}
-Don't use `Vue.use()`, `Vue.component()`, and globally, don't plug anything in Vue **inside** the `inject` function, dedicated to Nuxt injection. It will cause memory leak on server-side.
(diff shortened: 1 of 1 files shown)


content/en/docs/4.directory-structure/10.plugins.md
```diff
@@ -231,7 +231,7 @@ export const actions = {
 ```
 
 ::alert{type="warning"}
-Don't use `Vue.use()`, `Vue.component()`, and globally, don't plug anything in Vue **inside** this function, dedicated to Nuxt injection. It will cause memory leak on server-side.
+Don't use `Vue.use()`, `Vue.component()`, and globally, don't plug anything in Vue **inside** the `inject` function, dedicated to Nuxt injection. It will cause memory leak on server-side.
```
danielroe, line 234:
```suggestion
Don't use `Vue.use()`, `Vue.component()`, and don't make changes to the Vue prototype or global Vue object **inside** the function exported by your plugin. (It would cause a memory leak on server-side.) See also [global mixins](#global-mixins).
```

danielroe (approved):

PR #2030, opened 2022-03-08
docs: fix typo

azizurrahmaan:
I think `plugins which uses` should be `plugins which use` instead in the following line.
https://github.com/nuxt/nuxtjs.org/blob/cdcb2cea9221a30aefb518ef4bf8f936333c1170/content/en/docs/6.internals-glossary/1.context.md?plain=1#L163


danielroe (approved):
Thanks!

PR #2035, opened 2022-03-14
docs: recommend Volar instead of Vetur

sanscheese:
Volar is the recommended extension for [Vue 3 local setup](https://vuejs.org/guide/quick-start.html#with-build-tools) with VS Code.

I would think this should be the same for Nuxt 3?


sanscheese:
Ah sorry! Was looking at v2 docs. Nothing to see here.

the-simpsons-homer

PR #2036, opened 2022-03-16
Add content/pt/examples section translated

nazarepiedady:
In this pull request I am sending one more small part of the job I did, I did struggle
to find some ideal Portuguese equivalent for some Frontend technical words, 
so I want ask to you guys to review carefully the folders:

- `4.transitions/`
-  `6.loading/`
- `7.miscellaneous/`
- `8.middlewares/`
- `9.plugins/`

@smarroufin and @ChristopheCVB do you mind 😊?

### DIFF
--- a/content/pt/deployments/1.netlify.md
+++ b/content/pt/deployments/1.netlify.md
@@ -19,7 +19,7 @@ O centro do processo influencia o comando `nuxt generate`(<= v2.12) durante o de
 <div class="Promo__Video">
   <a href="https://explorers.netlify.com/learn/get-started-with-nuxt/nuxt-generate-and-deploy" target="_blank">
     <p class="Promo__Video__Icon">
-      Assista uma aula gratuita em <strong>Como fazer deploy de uma aplicação Nuxt no Netlify</strong> no Jamstack Explorers
+      Assista uma aula gratuita sobre <strong>Como desdobrar uma aplicação Nuxt no Netlify</strong> no Jamstack Explorers
     </p>
   </a>
 </div>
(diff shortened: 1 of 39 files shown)


smarroufin (approved):

content/pt/deployments/1.netlify.md
```diff
@@ -19,7 +19,7 @@ O centro do processo influencia o comando `nuxt generate`(<= v2.12) durante o de
 <div class="Promo__Video">
   <a href="https://explorers.netlify.com/learn/get-started-with-nuxt/nuxt-generate-and-deploy" target="_blank">
     <p class="Promo__Video__Icon">
-      Assista uma aula gratuita em <strong>Como fazer deploy de uma aplicação Nuxt no Netlify</strong> no Jamstack Explorers
+      Assista uma aula gratuita sobre <strong>Como desdobrar uma aplicação Nuxt no Netlify</strong> no Jamstack Explorers
```
promatik, line 22: @nazarepiedady first of all, thank you for your dedication 🙌
Do you use "desdobrar" often to refer to "deploy"?
Is it used in Angola? Because we don't use it in Portugal, the most common meaning of "desdobrar" is to "unfold".

By the way, Portuguese in Angola is VERY similar to Portuguese in Portugal, so I totally support the same version for pt-PT and pt-AO 👌

nazarepiedady: @promatik let me thank you for your observation and you really right the word "unfold" could be translated as "desdobrar" too.

I did choose to translate "deploy" to "desdobrar" because one of the meaning of this word is "to separate the all of something in two parts" that is in some sense what we do which consist in generate one copy of the project that will run on separated environment that we call "production".

However, I am open to suggestions that will improve the approach used in this version of the translation. 😊

PR #2037, opened 2022-03-20
docs: fix typo

YoonHan:
I think the semicolon might be inserted mistakenly. Please check out line number 76 in the following.

https://github.com/nuxt/nuxtjs.org/blob/ba319a8fc34160542076ffff1213c814831cac36/content/en/docs/3.features/8.nuxt-components.md?plain=1#L73-L90


smarroufin:
Hi @YoonHan 
You're right, this doesn't look correct.
Thanks for the feedback!

smarroufin (approved):

PR #2061, opened 2022-04-18
Fixing typo on 1.fetch.md

Flinty916:
Line 51: "falsy" to "false"


content/en/docs/7.components-glossary/1.fetch.md
```diff
@@ -48,7 +48,7 @@ You can access the Nuxt [context](/docs/internals-glossary/context) within the f
 - `fetchKey`: `String` or `Function` (defaults to the component scope ID or component name), a key (or a function that produces a unique key) that identifies the result of this component's fetch (available on Nuxt 2.15+) [More information available in original PR](https://github.com/nuxt/nuxt.js/pull/8466).
 - `fetchDelay`: `Integer` (default: `200`), set the minimum executing time in milliseconds (to avoid quick flashes)
 
-When `fetchOnServer` is falsy (`false` or returns `false`), `fetch` will be called only on client-side and `$fetchState.pending` will return `true` when server-rendering the component.
+When `fetchOnServer` is false (`false` or returns `false`), `fetch` will be called only on client-side and `$fetchState.pending` will return `true` when server-rendering the component.
```
danielroe, line 51:
```suggestion
When `fetchOnServer` is `false` or returns `false`, `fetch` will be called only on client-side and `$fetchState.pending` will return `true` when server-rendering the component.
```

danielroe:
I think this is preferable, but just FYI, `falsy` isn't a typo. It refers to any non-truthy value, such as undefined, zero, an empty string or null.

Atinux (approved):

PR #2081, opened 2022-05-10
chore(content): resync case studies

smarroufin:
Re-synchronize case studies


ChristopheCVB (approved):
PT translation is great!
Although, I see other languages have been changed in this PR, not sure why.

smarroufin:
@ChristopheCVB Those articles are not intended to be translated, so they should be the same across all languages. One file as been updated in PT because it has been "unsynced" at some point. I updated the EN folder and then copy pasted to all other folders.

@clemcode Shouldn't we ship that asap?

PR #2072, opened 2022-05-03
docs(pt): add `_collections` translated

nazarepiedady:
In this pull request, I am sending the `_collections` translated.

I remember that one day @smarroufin said that have the whole
`pt/docs/` folder translated would be enough to generate a new
website build putting the Portuguese version of the documentation
available, but I would like to ask you for a little 🙏 patience because
I am almost finished translating all files in the `pt/` folder 😊.

With this pull request, it will only be missing the following folders:

- `pt/announcements/`: it being translated.
- `pt/case-studies/`: it being translated.
- `pt/contribution-guide/`: the only file there is already translated.
- `pt/tutorials`: the translation has not started yet.

One more thing 😊,
@smarroufin what kind of contribution someone should do to be here [nuxt.org/teams](https://nuxtjs.org/teams) in the community section?

@smarroufin and @ChristopheCVB it is your turn.

PR #2099, opened 2022-06-18
docs(pt): add `content/pt/case-studies/github.md` translated

nazarepiedady:
Now, I start sending the `case-studies` folder to forward to the goal
of translating the whole website, at least the translatable parts. I intend
to start sending one file each time because in these ones I want a little 
more careful about the quality of the translations.

@smarroufin and @ChristopheCVB it is your turn now (we are back 😊).

### DIFF
--- a/content/en/docs/5.configuration-glossary/2.configuration-build.md
+++ b/content/en/docs/5.configuration-glossary/2.configuration-build.md
@@ -1,30 +1,30 @@
 ---
-title: The build Property
+title: A propriedade build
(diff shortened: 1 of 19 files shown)


nazarepiedady:
> Hi @nazarepiedady ! As is said previously, we don't want to translate those ones [#2072 (comment)](https://github.com/nuxt/nuxtjs.org/pull/2072#issuecomment-1115912297)

@smarroufin I did not remember about it, thanks.

PR #2107, opened 2022-07-07
Improve copy icon visibility

### LINKED ISSUE #2100: Copy button not visible enough in light mode
I changed it by applying `background-color: rgba(228,228,231,var(--tw-bg-opacity));` (which is equivalent to the `d-code-group-header-bg` class) to the `button` element and setting the stroke to `stroke: rgb(82, 82, 91);`

Before:
image

After:
image

hichemfantar:
Updated svg from the latest @docus/theme package.
Fixes #2100

### DIFF
--- a/components/atoms/icons/IconCopy.vue
+++ b/components/atoms/icons/IconCopy.vue
@@ -1,49 +1,10 @@
 <template>
-  <svg viewBox="0 0 20 16" fill="none" xmlns="http://www.w3.org/2000/svg">
-    <g filter="url(#filter0_d)">
-      <rect
-        x="4.60606"
-        y="0.606061"
-        width="9.45455"
-        height="9.45455"
-        rx="1.81818"
-        stroke="white"
-        stroke-width="1.21212"
-      />
-      <rect
-        x="9.93943"
-        y="5.93931"
-        width="9.45455"
-        height="9.45455"
-        rx="1.81818"
-        stroke="white"
-        stroke-width="1.21212"
-      />
-    </g>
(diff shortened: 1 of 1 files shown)


smarroufin (approved):
Thanks @hichemfantar ! 💚 
This is way better!

smarroufin:
I'm not fully sure about moving the v3 docs out of the source code repo but I would even say it's pretty unlikely.

hichemfantar:
Btw I noticed that some icons such as the copy icon are already provided by the @docus/theme package. Why not remove them from the project and update the package to the latest version and use the icons provided directly by the package theme.

smarroufin:
Yes, a lot of things can be improved or upgraded on repo. This is just that it has not been our focus since the website will change soon. Can't say much but in a way this will be resolved :slightly_smiling_face:

hichemfantar:
Understandable. What's the deal with Docus btw?
From what I read, It's supposed to be open source but all links that point to it are now dead.
Only way to read the source code is to go digging in my node_modules.

smarroufin:
There has been some rework on packages/orgs, it may be confusing :grimacing: 
Today, the `@docus/*` dependencies used by nuxtjs.org are kind of deprecated.
If you want to use markdown content, https://content.nuxtjs.org is the way with the v2 supporting nuxt v3.
If you want a docs theme based on content,  https://docus.dev (https://github.com/nuxtlabs/docus-theme-starter) is the way.

Hopping it's more clear, and it will be even clearer later :blush:

PR #2133, opened 2022-09-15
docs: add more explanation to server middleware section

a-toms:
Corrected a few typos and added a description of the client-server api pattern in Nuxt 2.

### DIFF
--- a/content/en/docs/5.configuration-glossary/28.configuration-servermiddleware.md
+++ b/content/en/docs/5.configuration-glossary/28.configuration-servermiddleware.md
@@ -16,7 +16,9 @@ Define server-side middleware.
 
 Nuxt internally creates a [connect](https://github.com/senchalabs/connect) instance that you can add your own custom middleware to. This allows us to register additional routes (typically `/api` routes) **without need for an external server**.
 
-This allows you to create a client api - server api pattern using nuxt alone. This means that you can use your client-side (e.g., a component) to call the api routes in your server middleware, i.e., your server-side routes. 
+This allows you to create a client api - server api pattern using nuxt alone. This means that you can use your client-side (e.g., a component) to call the api routes in your server middleware, i.e., your server-side middleware.
+
+One benefit of this pattern is that the server middleware exists on the server (like most middleware), not on the client. This means that you can handle environment variables and secrets in the server-side middleware, without exposing that information to the user.
 
 Because connect itself is a middleware, registered middleware will work with both `nuxt start` and also when used as a middleware with programmatic usages like [express-template](https://github.com/nuxt-community/express-template). Nuxt [Modules](/docs/directory-structure/modules) can also provide `serverMiddleware` using [this.addServerMiddleware()](/docs/internals-glossary/internals-module-container#addservermiddleware-middleware)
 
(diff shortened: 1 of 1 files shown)


content/en/docs/5.configuration-glossary/28.configuration-servermiddleware.md
```diff
@@ -29,12 +33,13 @@ Additional to them, we introduced a `prefix` option which defaults to `true`. It
 
 ## serverMiddleware vs middleware!
 
-Don't confuse it with [routes middleware](/docs/directory-structure/middleware) which are called before each route by Vue in Client Side or SSR. Middleware listed in the `serverMiddleware` property runs server-side **before** `vue-server-renderer` and can be used for server specific tasks like handling API requests or serving assets.
+Don't confuse it with [routes middleware](/docs/directory-structure/middleware) which are called before each route by Vue in Client Side or SSR. 
+Middleware listed in the `serverMiddleware` property runs server-side **before** `vue-server-renderer` and can be used for server specific tasks like handling API requests or serving assets.
```
danielroe, line 37:
```suggestion
Don't confuse it with [routes middleware](/docs/directory-structure/middleware) which are called before each route by Vue on the client and server.  Middleware listed in the `serverMiddleware` property runs on the server **before** `vue-server-renderer` and can be used for server-specific tasks like handling API requests or serving assets.
```

content/en/docs/5.configuration-glossary/28.configuration-servermiddleware.md
```diff
@@ -59,7 +64,7 @@ export default {
 ```
 
 ::alert{type="warning"}
-If you don't want middleware to register for all routes you have to use Object form with specific path, otherwise nuxt default handler won't work!
+If you don't want middleware to register for all routes, use Object form with specific path. If you don't do this, the nuxt default handler won't work!
```
danielroe, line 67:
```suggestion
If you don't want middleware to register for all routes, use the Object form with a specific path. If you don't do this, the nuxt default handler won't work!
```

content/en/docs/5.configuration-glossary/28.configuration-servermiddleware.md
```diff
@@ -16,6 +16,10 @@ Define server-side middleware.
 
 Nuxt internally creates a [connect](https://github.com/senchalabs/connect) instance that you can add your own custom middleware to. This allows us to register additional routes (typically `/api` routes) **without need for an external server**.
 
+This allows you to create a client api - server api pattern using nuxt alone. This means that you can use your client-side (e.g., a component) to call the api routes in your server-side middleware, i.e., your server middleware.
```
danielroe, line 19:
```suggestion
This allows you to create a client API/server API pattern using Nuxt alone. This means that from the browser (for example, within a Vue component) you can make a request to a route in your server middleware.
```

content/en/docs/5.configuration-glossary/28.configuration-servermiddleware.md
a-toms: 👍

danielroe (commented):
Thanks! ❤️

danielroe (approved):

PR #2137, opened 2022-09-24
docs(ja): fix mistranslations and improve fluency

NaokiHamada:
I have checked through the Japanese translation of the data-fetching page and fixed mistranslations in it.
Japanese fluency is also improved.


kazupon (approved):
Thank you for contribution!
Your translation is good!
Thanks!

PR #2141, opened 2022-09-27
fix: modules property example

MrZyr0:
The `~/` does not seems to work on my project. Replacing it by `@/` solve my problem


danielroe:
It should: https://stackblitz.com/edit/github-rrmkkt. Please feel free to open a discussion or issue (with a reproduction) if it doesn't work for you.

PR #2143, opened 2022-10-13
Removal of dist/ directory from .gignore mentioned

mbsc26:
A necessary step of removing dist/ folder from .gignore file is required for static deploment. This was not mentioned before, thus I added it.

### DIFF
--- a/content/en/docs/1.get-started/4.commands.md
+++ b/content/en/docs/1.get-started/4.commands.md
@@ -136,7 +136,7 @@ npm run generate
 ```
 ::
 
-Nuxt will create a `dist/` directory with everything inside ready to be deployed on a static hosting service.
+Nuxt will create a `dist/` directory with everything inside ready to be deployed on a static hosting service. Finally make sure to remove the `dist/` directory from your `.gitignore` file.
 
 As of Nuxt v2.13 there is a crawler installed that will now crawl your link tags and generate your routes when using the command `nuxt generate` based on those links.
 


manniL:
I am not sure if that is correct. As the generation should happen during the build process on the platform, the `dist` directory can stay in the `.gitignore` and should not be commited IMO

mbsc26:
@manniL brother there are 2 ways to do it: (1) either generate locally right before deploy in which case `dist/` folder has to be removed from `.gitignore` (2) Add generate as a post-build script which will generate `dist/` folder on the platform and thus removing `dist/` folder from `.gitgnore` is not required as you say. The docs mention method number (1) so the removal is necessary.

mbsc26:
@danielroe brother I am not committing the dist folder. I am using post-build script. But I just read the docs again. The article clearly misses out the point that `dist `should be ignored (as it says so for `.nuxt` folder under the Server Deployment section) and there is no mention of running of generate script on the target. In other words, the article suggests running of the script locally and committing dist

danielroe:
I just do not read it as saying the `dist` folder should be committed. It says:

> Nuxt will create a `dist/` directory with everything inside ready to be deployed on a static hosting service.

This page is a guide on what the _commands_ do, e.g. what `nuxt generate` does. There are specific deployment guides linked at the bottom of the page which will go into more details as to what to do for specific static providers.

Thank you for your time on this PR, but I think we should not add this suggestion.

mbsc26:
Yes, you are right no doubt but because the page specifically mentions to ignore `.nuxt` but does not metion to ignore `dist/`, this causes a confusion for learners. I will create the appropriate PR. Thanks a lot for your time as well.

mbsc26:
@danielroe Yes, you are right no doubt but because the page specifically mentions to ignore `.nuxt` but does not metion to ignore `dist/`, this causes a confusion for learners. I will create the appropriate PR. Thanks a lot for your time as well.

PR #2147, opened 2022-10-21
Update 11.configuration-generate.md - change "Ficheiro" to "Arquivo"

rodolfo-santos:
Here in Brazil, we use the word "Arquivo" instead of "Ficheiro", I am requesting to change the word to the Brazilian Portuguese standard.


ChristopheCVB:
This is `pt`, not `pt-BR`, so I think we'll stick to the pt-pt form

PR #2153, opened 2022-10-29
Added PNPM instructions for installations

n3-rd:
PNPM has been rising in popularity, so it's only fitting that it's also included


danielroe:
Perhaps we should say something about `--shamefully-hoist` being required?

n3-rd:
> Perhaps we should say something about `--shamefully-hoist` being required?

Yes, just learnt that today!
I'll update it.

PR #2161, opened 2022-11-16
Update banners for v3 stable release.

leerob:

PR #2162, opened 2022-11-16
fix: remove `beta` label from banner in `pt-PT` lang

MattFerreira18:
Recently has been pushed to production the banners updates of new stable version (PR #2161), but from pt-PT lang ins't removed the `beta` label.


ChristopheCVB (approved):

smarroufin (approved):
Thanks @MattFerreira18 💚

PR #2165, opened 2022-11-16
fix: changed AppLink to A tag, because is external link

vitordesousa:


danielroe:
apologies, `<AppLink>` is correct; it was just a deployment issue...

PR #2173, opened 2022-11-29
Update Docs, closes #925 and a lot of beginners problems

emanuele-em:
Very common problem with node version (also LTS versions) can be solved using node 16, that works good for this documentation, close issue


Atinux:
Closing as same reason as in nuxt/framwork since LTS is 18

PR #2176, opened 2022-12-02
Fix error in docs -> `nuxt` should be `nuxt dev`

louwers:
`nuxt` will just print documentation, it should be `nuxt dev`.


danielroe (commented):
you're absolutely right that that is correct for Nuxt 3, but this repo is for the Nuxt 2 docs. We need to make that clearer. You can find the Nuxt 3 docs in https://nuxt.com and https://github.com/nuxt/framework/tree/docs.

PR #2184, opened 2022-12-15
docs(partners): redirect to nuxt.com

clemcode:
This PR removes partners from nuxtjs.org and redirects previous URLs to https://nuxt.com/support/agencies


danielroe (approved):

danielroe:
🔥

PR #2192, opened 2023-01-02
docs: update stormkit instructions

svedova:
- Hybrid applications are detected automatically so there is no more need to switch the toggle. 
- Suggest using `redirects.json` instead of `stormkit.config.yml`


danielroe (approved):
Thanks! ❤️

Issue #636, opened 2020-09-02
Azure deplyoment is outdated


pi0:
- https://github.com/nuxt/create-nuxt-app/issues/563#issuecomment-663526975
- https://nuxtjs.org/faq/deployment-azure-portal/

The backend option is deprecated and removed from `create-nuxt-app`

PR #2201, opened 2023-01-16
docs: updated azure related names and steps (#636)

lextm:
While people should be recommended to host Nuxt apps on App Service (Linux) or simply a Linux container via AKS or Azure Container Apps, the related steps should be updated to avoid confusion.

I noticed that "The backend option is deprecated and removed from create-nuxt-app" was mentioned in the linked issue, but I don't fully understand its meaning and impact yet. Might send another pull request for that.

### DIFF
--- a/content/en/deployments/azure-portal.md
+++ b/content/en/deployments/azure-portal.md
@@ -1,16 +1,16 @@
 ---
 template: guide
-title: Azure Portal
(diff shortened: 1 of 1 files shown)


lextm:
Sounds like the latest approach is already documented, https://nitro.unjs.io/deploy/providers/azure, so no need for this pull request.

PR #2204, opened 2023-02-02
chore: redirect announcements to nuxt.com blog

clemcode:


content/en/_collections/navigations/footer.md
```diff
@@ -17,7 +17,7 @@ links:
   - title: 'Ecosystem'
     items:
       - title: 'Announcements'
-        to: '/announcements'
+        href: 'https://nuxt.com/blog'
```
danielroe, line 20: (This currently 404s, which I'm sure you're aware of - but just in case...)

PR #2207, opened 2023-02-09
docs: recommend either node 16.x or 14.x

wingzzzz:
Nodejs (version >= v18.0.0) is not supported!
See: https://github.com/nuxt/nuxt/issues/10844


danielroe (approved):
Thank you!

PR #2244, opened 2024-04-18
docs: using mjs on script

DeivisFelipe:
Change the extension of the nuxt file to .mjs


Atinux:
Not sure about this, this is still `.js` by looking at https://github.com/nuxt/nuxt/blob/2.x/distributions/nuxt/bin/nuxt.js

DeivisFelipe:
i will make all the process again with a empty project to make sure

DeivisFelipe:
Sorry i used the wrong docs, i am using Nuxt v3, but I read the v2

PR #2255, opened 2024-07-02
docs(pt): update the translated content inside `features` folder

nazarepiedady:
In this pull request, I am sending a big update to the Portuguese content inside `features` folder.
I found some misspelling errors in the original content in the process that'll be corrected soon when I finish the updates.

### DIFF
--- a/content/pt/docs/3.features/1.rendering-modes.md
+++ b/content/pt/docs/3.features/1.rendering-modes.md
@@ -1,31 +1,31 @@
 ---
-title: Renderização
+title: Modos de Interpretação
 category: features
 ---
 
-# Renderização
+# Modos de Interpretação
 
-## Sites Renderizados no Lado do Servidor e Sites Estáticos
+## Sítios Interpretados do Lado do Servidor e Sítios Estáticos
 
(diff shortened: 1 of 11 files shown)


nazarepiedady:
@smarroufin, @danielroe, @Atinux, @ChristopheCVB, @alexchopin, could you do the honours of reviewing the updates?

danielroe (approved):

danielroe:
As Nuxt v2 is now EOL, I think it might make sense not to devote more time to translating and maintaining the https://v2.nuxt.com website.

What do you think?

nazarepiedady:
@danielroe, I don't mind continuing, because my goal is to have a documentation with a uniform language with what I intend to do with the Nuxt 3 documentation.

I'm using the Nuxt 2 documentation in Portuguese as a way of preparing the ground for the Nuxt 3 documentation.

And since there will be projects that will continue to use Nuxt 2, it makes sense to update the documentation and leave it in excellent condition for those who will still need it.

I'll need your support in this endeavour, when I've finished the updates I'll let you know so you can archive the repository, if that's the next stage of the project.

PR #2256, opened 2024-07-16
docs(pt): update content inside `directory-structure` folder

nazarepiedady:
In this request to update the main branch of the repository, I'm sending the changes applied to the contents of the `directory-structure` folder in the Portuguese documentation.

I know that some may consider it a waste of time to update this material, but I believe that anyone who thinks this way has no real idea of how many existing projects still use Nuxt 2. 

Depending on the size of the project, there is a lot of bureaucracy and effort to overcome before you can start thinking about updating the technology stack.

My reasons go a little further than the points mentioned above, because I also want to standardise the language with the one I'll be using to translate the Nuxt 3 website.


nazarepiedady:
@Atinux, @danielroe, @ChristopheCVB, @smarroufin, @alexchopin. Could you do the honour of merging the updates?

